PR #16, opened 2020-10-20
update/asgardex-cosmos

junkai121:
implement asgardex-cosmos
- getAddress (derive path at index 0)
- getBalance
- getTransactions
- transfer
- deposit


packages/asgardex-cosmos/src/util.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export const isMsgMultiSend = (v: Msg): v is MsgMultiSend => 
+  (v as MsgMultiSend)?.inputs !== undefined && (v as MsgMultiSend)?.outputs !== undefined
+
+/**
+ * Get denom from Asset
+ */
+export const getDenom = (v: Asset): string => {
+  if (v === AssetAtom) return 'uatom'
+  if (v === AssetMuon) return 'umuon'
```
veado, line 23: It's safer to compare strings of assets here ^:

```
assetToString(v) === assetToString(AssetAtom)
// or
assetToString(v) === assetToString(AssetMuon)
```
In other cases `===` will force you to use the same object and it will fail if not ...

packages/asgardex-cosmos/src/util.ts
```diff
@@ -0,0 +1,34 @@
...
+  return v.symbol
+}
+
+/**
+ * Get Asset from denom
+ */
+export const getAsset = (v: string): Asset | null => {
+  if (v === 'uatom') return AssetAtom
+  if (v === 'umuon') return AssetMuon
```
veado, line 32: See previous message about comparing assets.

veado (changes requested):

veado (approved):
LGTM

Issue #28, opened 2020-10-21
FIX: Handle 0 balance for BinanceClient


thorchain-admin:
For Cosmos chains an address with 0 balance will return an error and cause the client to fall over. 

It should be caught (use `try{}catch(err){}`) and return a balance of 0 instead.

PR #30, opened 2020-10-21
Issue #28

junkai121:
- fix issue #28 
- update `lib` to xchainjs


packages/xchain-binance/src/client.ts
```diff
@@ -219,7 +219,7 @@ class Client implements BinanceClient, XChainClient {
         })
         .filter((balance) => !asset || balance.asset === asset)
     } catch (error) {
-      return Promise.reject(error)
+      return [] as Balances
```
veado, line 222: It might not a good idea to return an empty list of balances for ALL kind of errors. Is there a way to check the error to decide whether to return `[] as Balances` (for null balances accounts only) or `Promise.reject(error)` for all other errors?

veado (changes requested):

veado (approved):

PR #39, opened 2020-10-23
xchain-thorchain

junkai121:
add xchain-thorchain
fix issue #41


packages/xchain-thorchain/src/thor/types.ts
```diff
@@ -0,0 +1,59 @@
+import { BigSource } from 'big.js'
+
+import { Asset, THORChain } from '@xchainjs/xchain-util'
+import { PrivKey, Msg } from 'cosmos-client'
+import { BaseAccount, StdTx } from 'cosmos-client/x/auth'
+
+export const AssetThor: Asset = { chain: THORChain, symbol: 'THOR', ticker: 'THOR' }
```
veado, line 7: Don't we have RUNE here ^ (similar to [AssetRuneNative](https://gitlab.com/thorchain/asgardex-common/asgardex-util/-/blob/master/src/asset.ts#L120)
```
const AssetRune: Asset = { chain: THORChain, symbol: 'RUNE', ticker: 'RUNE' }
```
```

**Review by veado — CHANGES_REQUESTED** [review]
Since `xchain-thorchain` is very similar to `xchain-cosmos`, why not import and use similar stuff of `xchain-cosmos` (e.g. utils) here?

File: packages/xchain-cosmos/rollup.config.js [hunk 2]
```diff
@@ -33,5 +33,5 @@ export default {
     }),
     commonjs(),
   ],
-  external: ['readable-stream', 'buffer', 'stream', 'string_decoder'],
+  external: ['readable-stream', 'axios', 'buffer', 'cosmos-client', 'stream', 'string_decoder'],
```
veado, line 36: ^ `axios` can be removed, `crypto` is missing:

```
yarn build
...
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
crypto (imported by ../xchain-crypto/lib/index.es.js)
```

veado (approved):
Good work - just a small thing.

PR #42, opened 2020-10-26
xchain-cosmos

junkai121:
- update types for getTransaction and transfer
- update and move transaction parse to util, write tests for it.
- `yarn build`


veado (approved):
LGTM

Issue #25, opened 2020-10-21
Add `pre-commit` hook to lint packages


veado:
by using [husky](https://www.npmjs.com/package/husky) / [lint-staged](https://www.npmjs.com/package/lint-staged)

HelloKashif:
> by using [husky](https://www.npmjs.com/package/husky) / [lint-staged](https://www.npmjs.com/package/lint-staged)

Done 
![image](https://user-images.githubusercontent.com/9801383/97525196-a1597180-19e9-11eb-9bce-67215f3d8cb2.png)

PR #57, opened 2020-10-29
Github Workflow for test + precommit linting + fix linting issues

### LINKED ISSUE #23: Setup Github Actions to run tests for all packages
- Remove deprecated `.gitlab-ci.yml` in each package

HelloKashif:
Closes #23 #25

### DIFF
--- a/packages/xchain-bitcoin/src/client.ts
+++ b/packages/xchain-bitcoin/src/client.ts
@@ -217,16 +217,16 @@ class Client implements BitcoinClient, XChainClient {
     const offset = params?.offset ?? 0
 
     let totalCount = 0
-    let transactions: Tx[] = []
+    const transactions: Tx[] = []
     try {
       //Calling getAddress without limit/offset to get total count
       const dAddr = await blockChair.getAddress(this.nodeUrl, address, this.nodeApiKey)
       totalCount = dAddr[address].transactions.length
 
       const dashboardAddress = await blockChair.getAddress(this.nodeUrl, address, this.nodeApiKey, limit, offset)
-      let txList = dashboardAddress[address].transactions
+      const txList = dashboardAddress[address].transactions
 
-      for(let hash of txList){
+      for (const hash of txList) {
         const rawTx = (await blockChair.getTx(this.nodeUrl, hash, this.nodeApiKey))[hash]
         const tx: Tx = {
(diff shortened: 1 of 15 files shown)


stuartkuentzel (approved):

packages/xchain-bitcoin/src/client.ts
```diff
@@ -311,9 +311,9 @@ class Client implements BitcoinClient, XChainClient {
     }
     const calcdFees: Fees = {
       type: feesOptions.type,
-      fast: baseAmount(feesOptions.fast!.amount().multipliedBy(nextBlockFeeRate)),
-      average: baseAmount(feesOptions.average!.amount().multipliedBy(nextBlockFeeRate)),
-      fastest: baseAmount(feesOptions.fastest!.amount().multipliedBy(nextBlockFeeRate)),
+      fast: baseAmount(feesOptions.fast?.amount().multipliedBy(nextBlockFeeRate)),
+      average: baseAmount(feesOptions.average?.amount().multipliedBy(nextBlockFeeRate)),
+      fastest: baseAmount(feesOptions.fastest?.amount().multipliedBy(nextBlockFeeRate)),
```
veado, line 316: @HelloKashif ^ It will break if `feesOptions.fast` or `feesOptions.average` or `feesOptions.fastest` is undefined.

veado:
@HelloKashif Pls check GH actions again:

Screenshot from 2020-10-30 06-51-15

^ https://github.com/xchainjs/xchainjs-lib/actions

packages/xchain-bitcoin/src/client.ts
HelloKashif: @veado feesOptions is defined above so it shouldn't be undefined.
We can either move to ! mark instead of ?
OR
put a default value using ?? like
`feesOption.fast?.amount() ?? 5`

which one do you recommend?

veado: I would recommend to change `type Fees` as follow to avoid optional values.
```
type Fees = {
    type: FeeType
    fastest: BaseAmount
    fast: BaseAmount
    average: BaseAmount
};
```
If there is any chain, which provides one fee only, we will set same value for `average`, `fast`, `fastest`. It will be easier for frontends, too.

PR #70, opened 2020-11-04
Update fee handling

### LINKED ISSUE #65: Update `Fees`
Make values for  `fast` and `fastest` necessary (not optional). If there is any chain, which provides one fee only, we will set same value for `average`, `fast`, `fastest`.

### HAVE

```ts
export type Fees = {
  type: FeeType
  fastest?: BaseAmount
  fast?: BaseAmount
  average: BaseAmount
}
```

### WANT

```ts
type Fees = {
    type: FeeType
    fastest: BaseAmount
    fast: BaseAmount
    average: BaseAmount
}
```

### LINKED ISSUE #67: xchain-bitcoin: Add `getFeeRates`
Information about fee rates seems to be get lost while migrating `asgardex-bitcoin` into `xchain-bitcoin`. Let's bring it back by providing a `getFeeRates` method in `client.ts`

veado:
- [x] Make `average`, `fast`, `fastest` necessary for all clients
- [x] Update BTC fee handling to provide `feeRates` + `feeRatesWithMemo`

Closes #65 
Closes #67

### DIFF
--- a/packages/xchain-bitcoin/src/client.ts
+++ b/packages/xchain-bitcoin/src/client.ts
@@ -30,6 +30,7 @@ import { FeeData, FeeRate, FeeRates } from './types/client-types'
  */
 interface BitcoinClient {
   validateAddress(address: string): boolean
+  getFeesWithMemo(memo: string): Promise<Fees>
   getFeeRates(): Promise<FeeRates>
   getFeeRatesWithMemo(memo: string): Promise<FeeRates>
   scanUTXOs(): Promise<void>
@@ -255,7 +256,7 @@ class Client implements BitcoinClient, XChainClient {
   /**
    * Calculates fees based on fee rate and memo
    */
-  calcFee = (feeRate: FeeRate, memo?: string): BaseAmount => {
+  private calcFee = (feeRate: FeeRate, memo?: string): BaseAmount => {
     if (memo) {
       const OP_RETURN = Utils.compileMemo(memo)
       const vaultFee = Utils.getVaultFee(this.utxos, OP_RETURN, feeRate)
@@ -268,7 +269,7 @@ class Client implements BitcoinClient, XChainClient {
   /**
    * Returns rates and fees
    */
-  getFeeData = async (memo?: string): Promise<FeeData> => {
+  private getFeeData = async (memo?: string): Promise<FeeData> => {
     await this.scanUTXOs()
     if (this.utxos.length === 0) {
       throw new Error('No utxos to send')
--- a/packages/xchain-bitcoin/src/client.ts
+++ b/packages/xchain-bitcoin/src/client.ts
@@ -13,10 +13,12 @@ import {
   Balance,
   Network,
   Fees,
+  FeeOptionKey,
   XChainClientParams,
 } from '@xchainjs/xchain-client'
 import * as xchainCrypto from '@xchainjs/xchain-crypto'
-import { baseAmount, assetToString, AssetBTC } from '@xchainjs/xchain-util'
+import { baseAmount, assetToString, AssetBTC, BaseAmount } from '@xchainjs/xchain-util'
+import { FeeData, FeeRate, FeeRates } from './types/client-types'
 
 // https://blockchair.com/api/docs#link_300
 // const baseUrl = 'https://api.blockchair.com/bitcoin/'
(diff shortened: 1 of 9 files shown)


packages/xchain-bitcoin/src/client.ts
```diff
@@ -249,88 +252,101 @@ class Client implements BitcoinClient, XChainClient {
...
-  //   }
-  //   psbt.signAllInputs(btcKeys) // Sign all inputs
-  //   const tx = psbt.finalizeAllInputs().extractTransaction() // Finalise inputs, extract tx
-  //   const inputs = this.utxos.length // Add weight for each input sig
-  //   return tx.virtualSize() + inputs
-  // }
+  /**
+   * Calculates fees based on fee rate and memo
+   */
+  calcFee = (feeRate: FeeRate, memo?: string): BaseAmount => {
```
thatStrangeGuyThorchain, line 258: just as a suggestion:
maybe there is a sense to mark internal things as `private` to decrease public API ?

packages/xchain-bitcoin/src/client.ts
```diff
@@ -249,88 +252,101 @@ class Client implements BitcoinClient, XChainClient {
...
-      average: baseAmount(feesOptions.average?.amount().multipliedBy(nextBlockFeeRate)),
-      fastest: baseAmount(feesOptions.fastest?.amount().multipliedBy(nextBlockFeeRate)),
+
+    return { fees, rates }
+  }
+
+  /**
+   * Returns fees for transactions w/o a memo
+   */
+  getFees = async (): Promise<Fees> => {
```
thatStrangeGuyThorchain, line 298: hm... im really not sure that this is makes sense to separate `getFees` and `getFeesWithMemo` as they do the same (and if i understand correctly they WILL do the same all the time) so maybe we can set 
```ts
getFees = async (memo?: string): Promise<Fees>
``` 

the same thought about `getFeeRates/getFeeRatesWithMemo`

packages/xchain-cosmos/src/client.ts
```diff
@@ -214,14 +214,14 @@ class Client implements CosmosClient, XChainClient {
 
   // Need to be updated
   getFees = async (): Promise<Fees> => {
```
thatStrangeGuyThorchain, line 216: do we still need `getFees` to be async ?

packages/xchain-thorchain/src/client.ts
```diff
@@ -219,14 +219,14 @@ class Client implements ThorchainClient, XChainClient {
 
   // Need to be updated
   getFees = async (): Promise<Fees> => {
```
thatStrangeGuyThorchain, line 221: same question here

thatStrangeGuyThorchain (approved):
just few questions but its up to you

packages/xchain-bitcoin/src/client.ts
veado: It's because `interface` definition of `XChainClients` https://github.com/xchainjs/xchainjs-lib/blob/6831adb99d43292bc185d3194f3ad40fddf0e191/packages/xchain-client/src/types.ts#L91

Currently `BitcoinClient` is the only client which needs memos to calculate fees.

packages/xchain-cosmos/src/client.ts
veado: ^ Same reason as mentioned in https://github.com/xchainjs/xchainjs-lib/pull/70#discussion_r517481630

packages/xchain-thorchain/src/client.ts
veado: ^ Same reason as mentioned in https://github.com/xchainjs/xchainjs-lib/pull/70#discussion_r517481630

packages/xchain-bitcoin/src/client.ts
veado: > the same thought about getFeeRates/getFeeRatesWithMemo

Good point! In this case it does not break any interface of `XChainClients`! However, I would suggest to do it similar to `getFees` / `getFeesWithMemo` , just to be consistent.

packages/xchain-bitcoin/src/client.ts
veado: Good point!

PR #84, opened 2020-11-06
Mock Binance API

junkai121:


packages/xchain-binance/__tests__/client.test.ts
```diff
@@ -142,6 +249,8 @@ describe('BinanceClient Test', () => {
     const beforeTransfer = await client.getBalance()
     expect(beforeTransfer.length).toEqual(1)
 
+    await delay(1000) //delay before transaction
```
veado, line 252: Does we still need this delay by using mocks now?

packages/xchain-binance/__tests__/client.test.ts
```diff
@@ -164,8 +273,11 @@ describe('BinanceClient Test', () => {
     const beforeFreeze = await client.getBalance()
     expect(beforeFreeze.length).toEqual(1)
 
+    await delay(1000) //delay before transaction
```
veado, line 276: Same question here (and for all other delays we have in this test file): Is this delay still needed?

veado (commented):
Mocking of Binance API is looking good! Just a question ...

stuartkuentzel (approved):

veado (approved):

PR #90, opened 2020-11-09
Binance update client deps

thatStrangeGuyThorchain:


veado (approved):

veado (changes requested):
Code LGTM. Tests seems to be broken - please check it again.

PR #92, opened 2020-11-10
Fix derive path for xchain-bitcoin

junkai121:


packages/xchain-bitcoin/src/client.ts
```diff
@@ -52,6 +51,11 @@ class Client implements BitcoinClient, XChainClient {
   nodeUrl = ''
   nodeApiKey = ''
 
+  derive_path = {
+    mainnet: "84'/0'/0'/0/0",
+    testnet: "84'/1'/0'/0/0",
+  }
+
```
veado, line 58: Let keep it private: `private derive_path = {...}` or define `export const DERIVE_PATH = `{...}` outside from `Client`

packages/xchain-bitcoin/src/client.ts
```diff
@@ -128,13 +132,16 @@ class Client implements BitcoinClient, XChainClient {
...
-    const network = this.getNetwork() == 'testnet' ? Bitcoin.networks.testnet : Bitcoin.networks.bitcoin
-    const seed = BIP39.mnemonicToSeedSync(_phrase)
-    const wif = WIF.encode(network.wif, seed, true)
-    // TODO (@junkai121) Use `xchainCrypto.getSeed` while fixing it https://github.com/xchainjs/xchainjs-lib/issues/88
-    // const seed = xchainCrypto.getSeed(_phrase)
-    return Bitcoin.ECPair.fromWIF(wif, network)
+  private getBtcKeys(phrase: string): Bitcoin.ECPairInterface {
+    const isTestnet = this.getNetwork() === 'testnet'
+
+    const network = isTestnet ? Bitcoin.networks.testnet : Bitcoin.networks.bitcoin
```
veado, line 138: Since we do same thing in other methods, how about to extract it into `btcNetwork = (): Network => {...}`?

Same for `isTestnet`.

packages/xchain-bitcoin/src/client.ts
```diff
@@ -128,13 +132,16 @@ class Client implements BitcoinClient, XChainClient {
...
+    const derive_path = isTestnet ? this.derive_path.testnet : this.derive_path.mainnet
```
veado, line 139: How about to provide a `derivePath` method? If we make `derive_path` private, the users can still ask for the path.

veado (changes requested):
LGTM. Just few suggestions.

veado (approved):
P.S. Pls don't merge before tests have been fixed.

HelloKashif:
The failing tests have been fixed in this PR: https://github.com/xchainjs/xchainjs-lib/pull/98

Pr-98 already contains this PR's stuff so we can close this one and just merge 98

PR #109, opened 2020-11-23
[xchain-cosmos] Update deps + bump 0.4.1

veado:


junkai121 (approved):
LGTM

yarn.lock
```diff
@@ -900,6 +935,40 @@
     slash "^3.0.0"
     strip-ansi "^6.0.0"
 
+"@jest/core@^26.6.3":
```
thatStrangeGuyThorchain, line 938: duplicated too

thatStrangeGuyThorchain (commented):
left comments for few duplicated packages (not all). seems we should do a revision in a future

PR #122, opened 2020-12-01
Add `address derivation` for xchain-polkadot

junkai121:


packages/xchain-polkadot/src/client.ts
```diff
@@ -35,6 +35,7 @@ class Client implements PolkadotClient, XChainClient {
   private phrase = ''
   private address: Address = ''
   private api: ApiPromise | null = null
+  private derivationPath = "44//354//0//0//0'"
```
veado, line 38: Does `//` (double slash) in path mean `hard-derived` [as described in documentation](https://substrate.dev/docs/en/knowledgebase/integrate/subkey#putting-it-all-together) of Polkadot's Subkey Tool?

veado (approved):

packages/xchain-polkadot/src/client.ts
junkai121: Yes, right

PR #127, opened 2020-12-03
Update supported chains

### LINKED ISSUE #125: Update supported chains
in [const chains](https://github.com/xchainjs/xchainjs-lib/blob/6831adb99d43292bc185d3194f3ad40fddf0e191/packages/xchain-util/src/chain.const.ts#L15). Whenever a new chain will be supported, we have to update `chains` as well.

Remove deprecated comment about [All possible chains Thornode currently supports ...](https://github.com/xchainjs/xchainjs-lib/blob/6831adb99d43292bc185d3194f3ad40fddf0e191/packages/xchain-util/src/chain.const.ts#L13) to be Thorchain agnostic
```

### PR DESCRIPTION
**Pull request by junkai121**
Fix #125

### DIFF
--- a/packages/xchain-util/src/chain.const.ts
+++ b/packages/xchain-util/src/chain.const.ts
@@ -1,15 +1,14 @@
 /**
  * Chains
- *
- * Based on definition in Thorchain `common`
- * https://gitlab.com/thorchain/thornode/-/blob/master/common/chain.go#L15-18
  */
 export const BNBChain = 'BNB'
 export const BTCChain = 'BTC'
 export const ETHChain = 'ETH'
 export const THORChain = 'THOR'
+export const CosmosChain = 'GAIA'
+export const PolkadotChain = 'DOT'
 
 /**
- * All possible chains Thornode currently supports (or plan to support in near future)
+ * All possible chains XChainJS currently supports
  * */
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: packages/xchain-util/src/chain.const.ts [hunk 1]
```diff
@@ -1,15 +1,14 @@
...
- *
- * Based on definition in Thorchain `common`
- * https://gitlab.com/thorchain/thornode/-/blob/master/common/chain.go#L15-18
  */
 export const BNBChain = 'BNB'
 export const BTCChain = 'BTC'
 export const ETHChain = 'ETH'
 export const THORChain = 'THOR'
+export const CosmosChain = 'GAIA'
+export const PolkadotChain = 'POLKADOT'
```
veado, line 9: Don't we use `DOT` ([as discussed in Slack](https://thorchain.slack.com/archives/CPMV9FN80/p1606963852131100?thread_ts=1606909496.125400&cid=CPMV9FN80)) or do I miss something?

veado (commented):
Just a question...

packages/xchain-util/src/chain.const.ts
junkai121: We are now going with `POLKA`.`DOT` for polkadot. Just fixed and updated pr.

veado (approved):

PR #132, opened 2020-12-07
BTC Ledger Support

sarawut11:


packages/xchain-client/src/types.ts
```diff
@@ -56,6 +56,14 @@ export type TxParams = {
...
 }
 
+export type LedgerTxParams = {
+  asset?: Asset
+  amount: BaseAmount
+  sender: Address
+  recipient: Address
+  memo?: string // optional memo to pass
+}
+
```
veado, line 66:
```ts
export type LedgerTxParams = TxParams & {  sender: Address }
```

packages/xchain-bitcoin/src/client.ts
```diff
@@ -441,6 +445,62 @@ class Client implements BitcoinClient, XChainClient {
...
+    return {
+      utxos: this.utxos,
+      newTxHex: psbt.data.globalMap.unsignedTx.toBuffer().toString('hex'),
+    }
+  }
+
+  broadcastLedgerTx = async (txHex: string) => {
+    return await blockChair.broadcastTx(this.nodeUrl, txHex, this.nodeApiKey)
+  }
 }
```
veado, line 504: How about to rename it to `broadcastLedgerTx`, make it `private` and use it in `createTxForLedger` and `transfer`?

packages/xchain-bitcoin/src/client.ts
```diff
@@ -441,6 +445,62 @@ class Client implements BitcoinClient, XChainClient {
...
+        witnessUtxo: UTXO.witnessUtxo,
+      }),
+    )
+
+    // Outputs
+    psbt.addOutput({ address: recipient, value: amount.amount().toNumber() }) // Add output {address, value}
+    if (compiledMemo) {
+      // if memo exists
+      psbt.addOutput({ script: compiledMemo, value: 0 }) // Add OP_RETURN {script, value}
+    }
```
veado, line 493: Is there a way to extract all duplicated code into a function, which can be used in `createTxForLedger` and `transfer`

packages/xchain-bitcoin/src/client.ts
```diff
@@ -441,6 +445,62 @@ class Client implements BitcoinClient, XChainClient {
...
+
+    return {
+      utxos: this.utxos,
```
veado, line 496: [question] Why do we need to return all utxos?

veado (changes requested):

packages/xchain-bitcoin/src/client.ts
sarawut11: Those utxos are needed to do ledger sign.

packages/xchain-bitcoin/src/client.ts
sarawut11: createTxForLedger is just building tx and it can't be fully built cause ledger sign can't be handled here.
so the aim of this method returns necessary data to where ledger sign can be performed.
Thus, broadcastLedgerTx can't be a part of createTxForLedger

veado: Make sense. However, you can still in `transfer`. It's just about to remove duplication of code.

packages/xchain-bitcoin/src/client.ts
```diff
@@ -387,15 +391,17 @@ class Client implements BitcoinClient, XChainClient {
...
-    await this.scanUTXOs()
-    const balance = await this.getBalance()
+    sender,
+  }: TxParams & { feeRate: FeeRate; sender?: string }): Promise<Bitcoin.Psbt> => {
+    const address = sender || this.getAddress()
+    await this.scanUTXOs(address)
+    const balance = await this.getBalance(address)
     const btcBalance = balance.find((balance) => balance.asset.symbol === asset.symbol)
     if (!btcBalance) {
       throw new Error('No btcBalance found')
```
veado, line 407: Pls change in `buildTx` all 
```
throw new Error(...)
```
into
```
return Promise.reject(new Error(...))
```

packages/xchain-bitcoin/src/client.ts
```diff
@@ -428,17 +433,38 @@ class Client implements BitcoinClient, XChainClient {
...
+  createTxForLedger = async (params: LedgerTxParams & { feeRate: FeeRate }): Promise<Utils.LedgerTxInfo> => {
+    const psbt = await this.buildTx(params)
+
+    return {
+      utxos: this.utxos,
+      newTxHex: psbt.data.globalMap.unsignedTx.toBuffer().toString('hex'),
+    }
+  }
+
+  broadcastTx = async (txHex: string): Promise<TxHash> => {
```
veado, line 467: Pls add `broadcastTx` to `interface BitcoinClient`

veado (changes requested):

veado (approved):

PR #141, opened 2020-12-14
Update `getTransactionData`, `getTransactions` (xchain-thorchain)

junkai121:


packages/xchain-thorchain/src/util.ts
```diff
@@ -14,5 +18,86 @@ export const getDenom = (v: Asset): string => {
...
+ */
+export const parseAmountString = (v: string): TransferAmount | undefined => {
+  try {
+    const value = parseInt(v)
+    const denom = v.replace(value.toString(), '')
+    return { value, denom }
+  } catch (e) {
+    return undefined
+  }
+}
```
veado, line 55: Using `parseInt` have some limitations in JS. We can avoid it by parsing `value`, `denom` using `RegEx`. To store values more type safety we can also use updated type `TransferAmount` (see previous comment).

packages/xchain-thorchain/src/types.ts
```diff
@@ -1,4 +1,9 @@
 import { Asset } from '@xchainjs/xchain-util'
 
+export type TransferAmount = {
+  value: number
+  denom: string
+}
+
```
veado, line 7: A more type-safety way:
```ts
export type TransferAmount = {
  value: BaseAmount
  asset: Asset
}
```

veado (changes requested):

veado (approved):

PR #144, opened 2020-12-17
Fix `getBalance`, `getTransactions`, `getTransactionData`

junkai121:


packages/xchain-ethereum/src/client.ts
```diff
@@ -216,15 +212,13 @@ export default class Client implements XChainClient, EthereumClient {
   }
 
   // Returns balance of address
-  getBalance = async (addressParam?: string): Promise<Balance[]> => {
-    const address = addressParam || this.getAddress()
-
+  getBalance = async (address: string = this.getAddress()): Promise<Balance[]> => {
```
veado, line 215: `this.getAddress()` [might throw](https://github.com/xchainjs/xchainjs-lib/blob/890880c13e7c26d523c9d0dc5f58fc74b604baab/packages/xchain-ethereum/src/client.ts#L118) - don't use it as default value for a function parameter

veado (changes requested):

veado (approved):

PR #146, opened 2020-12-18
Update xchain-thorchain (prepare bump @0.6.1)

junkai121:
### Update

- Add `setClientUrl`
- Add `getDefaultClientUrl`

### Fix

- Fix client URL for multichain testnet (`https://testnet.thornode.thorchain.info`)

### DIFF
--- a/packages/xchain-thorchain/src/client.ts
+++ b/packages/xchain-thorchain/src/client.ts
@@ -56,7 +56,7 @@ class Client implements ThorchainClient, XChainClient {
 
   constructor({ network = 'testnet', phrase, clientUrl }: XChainClientParams & { clientUrl?: ClientUrl }) {
     this.network = network
-    this.clientUrl = clientUrl ? clientUrl : this.getDefaultClientUrl()
+    this.clientUrl = clientUrl || this.getDefaultClientUrl()
     this.thorClient = this.getNewThorClient()
 
     if (phrase) this.setPhrase(phrase)
--- a/packages/xchain-thorchain/src/client.ts
+++ b/packages/xchain-thorchain/src/client.ts
@@ -70,6 +70,7 @@ class Client implements ThorchainClient, XChainClient {
   setNetwork = (network: Network): XChainClient => {
     if (this.network != network) {
       this.network = network
+      this.clientUrl = this.getDefaultClientUrl()
       this.thorClient = this.getNewThorClient()
       this.address = ''
     }
--- a/packages/xchain-thorchain/src/client.ts
+++ b/packages/xchain-thorchain/src/client.ts
@@ -19,7 +19,7 @@ import { PrivKey, codec, Msg, AccAddress } from 'cosmos-client'
 import { StdTx } from 'cosmos-client/x/auth'
 import { MsgSend, MsgMultiSend } from 'cosmos-client/x/bank'
 
-import { AssetRune, DepositParam } from './types'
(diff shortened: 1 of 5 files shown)


packages/xchain-thorchain/src/client.ts
```diff
@@ -68,14 +68,12 @@ class Client implements ThorchainClient, XChainClient {
...
-    this.thorClient = new CosmosSDKClient({
-      server: this.getClientUrl(),
-      chainId: this.getChainId(),
-      prefix: this.getPrefix(),
-      derive_path: this.derive_path,
-    })
-    this.address = ''
+    if (this.network != network) {
+      this.network = network
+      this.clientUrl = this.getDefaultClientUrl()
```
veado, line 73: ^ That would deprecate/override `clientUrl`, which may set using `setClientUrl` before.

packages/xchain-thorchain/src/client.ts
```diff
@@ -84,8 +82,26 @@ class Client implements ThorchainClient, XChainClient {
     return this.network
   }
 
+  setClientUrl = (clientUrl: string): void => {
```
veado, line 85: How about: `clientUrl: string` -> `{testnetClientUrl: string, mainnetClientUrl: string}`

packages/xchain-thorchain/src/client.ts
```diff
@@ -51,12 +55,8 @@ class Client implements ThorchainClient, XChainClient {
 
   constructor({ network = 'testnet', phrase }: XChainClientParams) {
```
veado, line 56: How about to provide `clienUrl` for `test|mainnet` as optional parameters? 
- `testnetClientUrl: string`
- `mainnetClientUrl: string`

packages/xchain-thorchain/src/client.ts
```diff
@@ -84,8 +82,26 @@ class Client implements ThorchainClient, XChainClient {
...
+    this.clientUrl = clientUrl
+    this.thorClient = this.getNewThorClient()
+  }
+
   getClientUrl = (): string => {
-    return this.network === 'testnet' ? 'http://18.198.92.45:1317' : 'http://138.68.125.107:1317'
+    return this.clientUrl
+  }
+
+  getDefaultClientUrl = (): string => {
```
veado, line 94: In addition to previous comment, return of `{testnetClientUrl: string, mainnetClientUrl: string}` is needed then. 

And you may need to have another function `getClientUrlByNetwork = (network: Network): string`.

packages/xchain-thorchain/src/client.ts
```diff
@@ -84,8 +82,26 @@ class Client implements ThorchainClient, XChainClient {
     return this.network
   }
 
+  setClientUrl = (clientUrl: string): void => {
+    this.clientUrl = clientUrl
+    this.thorClient = this.getNewThorClient()
+  }
+
   getClientUrl = (): string => {
```
veado, line 90: In addition to previous comment ^
`getClientUrl = (): {testnetClientUrl: string, mainnetClientUrl: string}`

veado (changes requested):

packages/xchain-thorchain/src/client.ts
junkai121: What do you think of following approach? @veado  It returns the client url of current network...
```
  getClientUrl = (): string => {
    return this.getClientUrlByNetwork(this.network)
  }
```

veado (approved):

Issue #118, opened 2020-11-29
Extract `validateAddress` into `XChainClient` interface 


veado:
- Add `validateAddress(address: Address): boolean` to `XChainClient
- Remove `validateAddress` from [BitcoinClient](https://github.com/xchainjs/xchainjs-lib/blob/master/packages/xchain-bitcoin/src/client.ts#L33), [BinanceClient](https://github.com/xchainjs/xchainjs-lib/blob/master/packages/xchain-binance/src/client.ts#L33), [CosmosClient](https://github.com/xchainjs/xchainjs-lib/blob/master/packages/xchain-cosmos/src/client.ts#L30), [ThorchainClient](https://github.com/xchainjs/xchainjs-lib/blob/master/packages/xchain-thorchain/src/client.ts#L30)
- Implement `validateAddress` to `asgardex-polkadot`

junkai121:
Resolved by #149

PR #149, opened 2020-12-22
Extract `validateAddress` into `XChainClient` interface

junkai121:
#118


packages/xchain-polkadot/src/client.ts
```diff
@@ -130,6 +131,15 @@ class Client implements PolkadotClient, XChainClient {
     }
   }
 
+  validateAddress = (address: string): boolean => {
```
veado, line 134: Could you add following link into a comment, pls?

https://polkadot.js.org/docs/util-crypto/examples/validate-address

It will help us and others to understand why we validate an address like this ^

veado (approved):
LGTM!

PR #157, opened 2020-12-24
move getDefaultFees from client to utils

sarawut11:


veado (approved):

veado (changes requested):
Code changes are looking good! Pls update CHANGELOG of each lib with something like this (it might be different for all libs):

```
# v.x.x.x (2020-XX-XX)

### Breaking change

- Extract `getDefaultFees` from `Client` to `utils`


```

veado (approved):

PR #160, opened 2020-12-30
Update `xchain-cosmos`

junkai121:
Update comments for xchain-cosmos.


packages/xchain-cosmos/src/client.ts
```diff
@@ -99,6 +177,15 @@ class Client implements CosmosClient, XChainClient {
     return this.getAddress()
   }
 
+  /**
+   * @private
+   * Get private key.
+   *
+   * @returns {PrivKey} The privkey generated from the given phrase
```
veado, line 184: [nitpick] `privkey` -> `private key`

veado (approved):

PR #170, opened 2021-01-07
xchain-thorcahin/getExplorerUrl update based on network

### LINKED ISSUE #152: [xchain-thorchain] Make getExplorerUrl return url based on network
now return value for ThorChain client's `getExplorerUrl` [is hardcoded.](https://github.com/xchainjs/xchainjs-lib/blob/master/packages/xchain-thorchain/src/client.ts#L127).
it should be based on current network

sarawut11:
closes #152

### DIFF
--- a/packages/xchain-thorchain/src/client.ts
+++ b/packages/xchain-thorchain/src/client.ts
@@ -207,6 +207,9 @@ class Client implements ThorchainClient, XChainClient {
    * @returns {string} The explorer url.
    */
   getExplorerUrl = (): string => {
+    if (this.network === 'testnet') {
+      return 'https://testnet.thorchain.net'
+    }
     return 'https://thorchain.net'
   }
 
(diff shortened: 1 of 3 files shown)


packages/xchain-thorchain/src/client.ts
```diff
@@ -38,6 +38,9 @@ export interface ThorchainClient {
   getDefaultClientUrl(): ClientUrl
   getClientUrlByNetwork(network: Network): string
   setClientUrl(clientUrl: ClientUrl): void
+  getDefaultExplorerUrl(): ExplorerUrl
+  getExplorerUrlByNetwork(network: Network): string
+  setExplorerUrl(explorerUrl: ExplorerUrl): void
```
veado, line 43: Those methods don't need to be public. Make it private and remove it from interface. Do the same with `getDefaultClientUrl`, `getClientUrlByNetwork`, `setClientUrl`

veado (changes requested):
Some notes about current interface. Pls update CHANGELOG.

veado (approved):

PR #172, opened 2021-01-07
update getTransactions to get txes in right order(xchain-thorchain)

sarawut11:


sarawut11:
This will be handled in another way. Will integrate tendermint, one of its rpcs support get transactions in asc/desc

PR #181, opened 2021-01-12
Move fromJson outside of MsgNativeTx

### LINKED ISSUE #179: [xchain-thorcain] TypeError: Class constructor Msg cannot be invoked without 'new'
using of [static fromJson](https://github.com/xchainjs/xchainjs-lib/blob/master/packages/xchain-thorchain/src/types/messages.ts#L29)

leads error 

``` TypeError: Class constructor Msg cannot be invoked without 'new'```
image

some thoughts:

i've tried to move `fromJson` method outside of class constructor at all and seems it works well.

junkai121:
Fix #179

### DIFF
--- a/packages/xchain-thorchain/CHANGELOG.md
+++ b/packages/xchain-thorchain/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - Fix getTransactions pagination issue #168
 
+### Breaking change
+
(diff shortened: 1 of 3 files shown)


veado (approved):
LGTM!

Please update CHANGELOG +  bump / publish xchain-thorchain@0.8.0

PR #184, opened 2021-01-14
Add `getPrefix`

junkai121:


veado (changes requested):
Code looks good in general, but I would suggest to add a `getPrefix` into `utils` module of each xchain-* client package (not into `xchain-util`). Because xchain-* clients will needed a dependency to `xchain-util` now (not all have it yet). Also it would import code and information into a xchain-* clients, we don't need there. We just need a prefix for the client's addresses, but not a check for all other chains.

veado (approved):
LGTM!

PR #188, opened 2021-01-20
Ethereum Fixes

### LINKED ISSUE #187: Ethereum Fixes
# Ethereum Issues

* Client needs to be agnostic and unconcerned with THORChain. Remove all aspects to do with "vault", eg `vaultTx` `setVault`
* Remove `normalTx`, replace with `transfer()` where the asset is specified, Ethereum is `0x0`, assets is passed in

ERC20
* If ERC-20, then must call the ERC20 ABI (hard-code into `/abi` and import) but still be called using `transfer(asset, amount, recipient)`
* Must have a hard-coded `isApproved(spender, sender, amount)` which checks to see if approved, return true/false, 
* Must have hard-coded `approve()` to do an ERC-20 infinite (uint(-1)) approval

Dealing with Smart Contracts
* Must be generalised using `call(abi, address, function, [params])`, where necessary is passed in
* Deposit to THORChain is a Contract Call, thus would be `call(router-abi, router-address, "deposit", [params])`

junkai121:
Fix #187


packages/xchain-ethereum/src/client.ts
```diff
@@ -717,21 +660,22 @@ export default class Client implements XChainClient, EthereumClient {
    * @throws {"Invalid Asset Address"}
    * Thrown if the given asset address is invalid.
    */
-  erc20Tx = async ({ assetAddress, recipient, amount, overrides }: Erc20TxOpts): Promise<TransactionResponse> => {
+  erc20Transfer = async ({ assetAddress, recipient, amount, overrides }: Erc20TxOpts): Promise<TxHash> => {
```
thorchain-admin, line 663: Do not have a separate public function for a ERC20-tx. The upstream wallet should not care if transferring ETH or ERC20, it is this client's job to detect which asset and proceed accordingly. 

Simply have one master `transfer()` function.

packages/xchain-ethereum/src/client.ts
```diff
@@ -57,16 +48,18 @@ import {
...
 export interface EthereumClient {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   call<T>(asset: Address, abi: ethers.ContractInterface, func: string, params: Array<any>): Promise<T>
-  vaultTx(opts: VaultTxOpts): Promise<TransactionResponse>
-  normalTx(opts: NormalTxOpts): Promise<TransactionResponse>
-  erc20Tx(opts: Erc20TxOpts): Promise<TransactionResponse>
+
+  isApproved(spender: Address, sender: Address, amount: BaseAmount): Promise<boolean>
+  approve(spender: Address, sender: Address, amount?: BaseAmount): Promise<TxHash>
+  erc20Transfer(opts: Erc20TxOpts): Promise<TxHash>
```
thorchain-admin, line 62: do not have this, use the `transfer()` function and handle logic inside of it

packages/xchain-ethereum/src/client.ts
```diff
@@ -57,16 +48,18 @@ import {
...
 
   estimateGasNormalTx(params: NormalTxOpts): Promise<BaseAmount>
   estimateGasERC20Tx(params: Erc20TxOpts): Promise<BaseAmount>
```
thorchain-admin, line 65: Do not have this, have a single `estimateGas(opts)` and return the correct gas amount whether ETH or ERC20

packages/xchain-ethereum/src/client.ts
thorchain-admin, line 663: Should be private `_transferERC20()`

packages/xchain-ethereum/src/types/client-types.ts
```diff
@@ -18,40 +18,21 @@ export type ExplorerUrl = {
...
 
-export type VaultTxOpts = {
-  address: Address
-  amount: BaseAmount
-  memo: string
-  overrides?: ContractTxOverrides
-}
-
-export type NormalTxOpts = {
+export type EstimateGasOpts = {
```
davidp94, line 21: I think you might need to add an optional `from` if you'd like to query the gas estimate depending of the sender; without having him to pass his private key everytime;

The `from` is important because if the recipient is a smart contract that behaves differently depending of the `sender`, his account balance (and the smart contract state), it will cost different gas.

davidp94: On Contract instance:

```
// https://docs.ethers.io/v5/api/contract/contract/#contract-estimateGas
overrides = { from: '0x....' }
contract.estimateGas.METHOD_NAME( ...args [ , overrides ] )
```

packages/xchain-ethereum/src/client.ts
```diff
@@ -576,67 +475,109 @@ export default class Client implements XChainClient, EthereumClient {
...
-   * @returns {TransactionResponse}  The vault transaction result.
+   * @param {Address} spender The spender address.
+   * @param {Address} sender The sender address.
+   * @param {BaseAmount} amount The amount of token.
+   * @returns {boolean} `true` or `false`.
    */
-  vaultTx = async ({ address, amount, memo, overrides }: VaultTxOpts): Promise<TransactionResponse> => {
+  private isApproved = async (spender: Address, sender: Address, amount: BaseAmount): Promise<boolean> => {
     try {
       const txAmount = parseEther(baseToAsset(amount).amount().toFormat())
```
veado, line 492: [question] Is it safe to use [Bignumber's `toFormat()`](https://mikemcl.github.io/bignumber.js/#toFor) here and at other places? Does [Bignumber's `toString()`](https://mikemcl.github.io/bignumber.js/#toS) does not work?

junkai121: `toString()` will parse BigNumber to `1e-18` which is invalid decimal value.

packages/xchain-ethereum/src/types/client-types.ts
junkai121: Thank you!

PR #198, opened 2021-01-29
Add Litecoin support

HelloKashif:
Few points to note:

- Blockchair API currently does not support LTC testnet so we migrated to using sochain API. 
- Since Sochain API does not provide fee rate data `getFees` method returns a fixed value of `1` 
- Sochain does not have a unified tx api (we need to call multiple apis to get unspent/spent txs) so the pagination/offset feature is not supported
- getTransactions only returns latest 50 txs


packages/xchain-litecoin/src/utils.ts
```diff
@@ -0,0 +1,347 @@
...
+
+/**
+ * Get the vault transaction fee.
+ *
+ * @param {UTXOs} inputs The UTXOs.
+ * @param {Buffer} data The compiled memo.
+ * @param {FeeRate} feeRate The fee rate.
+ * @returns {number} The fee amount.
+ */
+export function getVaultFee(inputs: UTXOs, data: Buffer, feeRate: FeeRate): number {
```
thorchain-admin, line 41: The language of "vault" and "normal" was deprecated when we moved from asgardex to xchainjs. XChainjs is agnostic and shouldn't have the concept of "vault".

The only difference between the two is the presence of the memo.

Request replace both functions. with a single "getFee()`, where the memo is optional

PR #211, opened 2021-02-03
[xchain-thorchain] get transaction history from the latest ones

### LINKED ISSUE #173: [xchain-thorchain] get transaction history from the latest ones
We are using cosmos rpc (https://cosmos.network/rpc/v0.37.9) and it returns old transactions first. (ascending order)
https://testnet.thornode.thorchain.info/txs?message.sender=tthor13gym97tmw3axj3hpewdggy2cr288d3qffr8skg&page=1

But if we use tendermint rpc(https://docs.tendermint.com/master/rpc/#/Info/tx_search), there is order_by option.

We can get the node ip address using https://testnet.thornode.thorchain.info/node_info
And we can use it with 26657 port (it's Tendermint RPC endpoint)

http://18.198.92.45:26657/tx_search?query=%22message.sender=%27tthor13gym97tmw3axj3hpewdggy2cr288d3qffr8skg%27%22

junkai121:
Fix #173 
prepare bump xchain-cosmos@0.8.0, xchain-thorchain@0.10.0


packages/xchain-thorchain/src/client.ts
```diff
@@ -389,9 +397,14 @@ class Client implements ThorchainClient, XChainClient {
         txMaxHeight,
       })
 
+      const txs: Txs = []
+      for (const tx of txHistory.txs) {
+        txs.push(await this.getTransactionData(tx.hash))
```
veado, line 402: To avoid many requests by default, set default values of [offset=0](https://github.com/xchainjs/xchainjs-lib/blob/86c13c20b4538ab14aadb09d1bf35d58154dd694/packages/xchain-thorchain/src/client.ts#L381) / [limit=10](https://github.com/xchainjs/xchainjs-lib/blob/86c13c20b4538ab14aadb09d1bf35d58154dd694/packages/xchain-thorchain/src/client.ts#L382) before.

veado (changes requested):
Code LGTM! One thought ...

veado (approved):

PR #227, opened 2021-02-15
Update Bitcoin Cash

junkai121:
- add `scanUTXOs`, `transfer`, `fee`

### DIFF
--- a/packages/xchain-bitcoincash/src/utils.ts
+++ b/packages/xchain-bitcoincash/src/utils.ts
@@ -136,7 +136,7 @@ export const validateAddress = (address: string, network: Network): boolean => {
  * @returns {Array<UTXO>} The UTXOs of the given address.
  */
 export const scanUTXOs = async (clientUrl: string, address: Address): Promise<bitcash.Transaction.UnspentOutput[]> => {
-  const unspents = await getUnspentTransactions(clientUrl, address)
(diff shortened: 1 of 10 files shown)


packages/xchain-bitcoincash/src/utils.ts
```diff
@@ -110,3 +112,143 @@ export const parseTransaction = (tx: Transaction): Tx => {
...
+}
+/**
+ * Scan UTXOs from sochain.
+ *
+ * @param {string} clientUrl sochain Node URL.
+ * @param {Address} address
+ * @returns {Array<UTXO>} The UTXOs of the given address.
+ */
+export const scanUTXOs = async (clientUrl: string, address: Address): Promise<bitcash.Transaction.UnspentOutput[]> => {
+  const unspents = await getUnspentTransactions({ clientUrl, address })
```
veado, line 139: What happened if `getUnspentTransactions` failed?

veado (changes requested):

packages/xchain-bitcoincash/src/utils.ts
junkai121: If it's failed, it will throw an error, Let me add try...catch for the API requests.

veado (approved):
:+1:

PR #238, opened 2021-02-19
fix getTransaction in ETH

sarawut11:


veado (changes requested):
Code LGTM! Please update / fix tests.

veado (approved):

PR #251, opened 2021-02-24
Fix `getTransactions`, `DECIMAL`

junkai121:
- fix `getTransactions` - sort from latest.
- fix Decimals
- prepare bump @0.11.1


veado (approved):
LGTM - thx!

PR #265, opened 2021-03-02
update getTransactionData type

### LINKED ISSUE #264: getTransactionData type issue
To get ERC20 token tx, it needs assetAddress but the xchain-lib interface getTransactionData has only txid as a param so it's not allowed to call getTransactionData with assetAddress.

sarawut11:
closes #264

### DIFF
--- a/packages/xchain-client/src/types.ts
+++ b/packages/xchain-client/src/types.ts
@@ -96,7 +96,7 @@ export interface XChainClient {
 
   getTransactions(params?: TxHistoryParams): Promise<TxsPage>
 
-  getTransactionData(txId: string, assetAddress?: string): Promise<Tx>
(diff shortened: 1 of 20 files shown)


packages/xchain-client/src/types.ts
```diff
@@ -96,7 +96,7 @@ export interface XChainClient {
 
   getTransactions(params?: TxHistoryParams): Promise<TxsPage>
 
-  getTransactionData(txId: string): Promise<Tx>
+  getTransactionData(txId: string, assetAddress?: string): Promise<Tx>
```
veado, line 99: [nitpick] `assetAddress?: string` -> `assetAddress?: Address`

veado (changes requested):

veado (approved):

PR #270, opened 2021-03-05
Update gasLimit estimation for xchain-ethereum

junkai121:
- update gas limit estimation logic.
- add `estimateCall`

### DIFF
--- a/packages/xchain-ethereum/CHANGELOG.md
+++ b/packages/xchain-ethereum/CHANGELOG.md
@@ -1,3 +1,13 @@
+# v.x.x.x
+
+### Update
+
+- Add `estimateCall` to estimate gaslimit for `call` function.
+
+### Fix
+
+- Fix gas limit estimation logic.
+
 # v.0.13.2 (2021-03-03)
 
 ### Update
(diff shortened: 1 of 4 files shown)


packages/xchain-ethereum/CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+# v.x.x.x
+
+### Update
+
+- Add `estimateCall` to estimate gaslimit for `call` function.
+
+### Fix
+
+- Fix gas limit estimation logic.
```
veado, line 9: Breaking change (types of `estimateGasLimit()` + `FeesWithGasPricesAndLimits` have been changed)

packages/xchain-ethereum/src/client.ts
```diff
@@ -669,15 +692,13 @@ export default class Client implements XChainClient, EthereumClient {
 
         estimate = await contract.estimateGas.transfer(recipient, txAmount, {
           from: this.getAddress(),
-          gasPrice: gasPriceBN,
```
veado, line 672: Why don't we need a value for `gasPrice` anymore?

veado (changes requested):

packages/xchain-ethereum/src/client.ts
junkai121: We don't need gasPrice for gas limit estimation.

veado (approved):

PR #275, opened 2021-03-09
Fix xchain-ethereum

junkai121:
- export Client.`estimateGasPrices`, Util.`getFee`
- fix Util.`getTokenAddress`
- save `etherscanApiKey` in Client

This pr is for fee handling using `estimateCall` on ASGDX.

and this will close #274

### DIFF
--- a/packages/xchain-ethereum/src/utils.ts
+++ b/packages/xchain-ethereum/src/utils.ts
@@ -92,7 +92,7 @@ export const validateAddress = (address: Address): boolean => {
  */
 export const getTokenAddress = (asset: Asset | null): string | null => {
   if (asset) {
-    const assetAddress = asset.symbol.slice(asset.ticker.length + 1)
+    const assetAddress = asset.symbol.slice(asset.ticker.length + 1).toLowerCase()
     if (validateAddress(assetAddress)) {
       return assetAddress
     }
--- a/packages/xchain-ethereum/src/utils.ts
+++ b/packages/xchain-ethereum/src/utils.ts
@@ -90,15 +90,12 @@ export const validateAddress = (address: Address): boolean => {
  * @param {Asset} asset
  * @returns {string|null} The token address.
  */
-export const getTokenAddress = (asset: Asset | null): string | null => {
-  if (asset) {
-    const assetAddress = asset.symbol.slice(asset.ticker.length + 1).toLowerCase()
-    if (validateAddress(assetAddress)) {
-      return assetAddress
-    }
+export const getTokenAddress = (asset: Asset): string | null => {
+  try {
+    return ethers.utils.getAddress(asset.symbol.slice(asset.ticker.length + 1).substr(2))
+  } catch (err) {
+    return null
   }
-
-  return null
 }
 
 /**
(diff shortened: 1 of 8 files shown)


packages/xchain-ethereum/src/utils.ts
```diff
@@ -92,7 +92,7 @@ export const validateAddress = (address: Address): boolean => {
  */
 export const getTokenAddress = (asset: Asset | null): string | null => {
```
veado, line 93: Can we change `(asset: Asset | null)` to `(asset: Asset)` for parameters of  `getTokenAddress`? _It will be a breaking change then (just for CHANGELOG) ..._

packages/xchain-ethereum/src/utils.ts
```diff
@@ -92,7 +92,7 @@ export const validateAddress = (address: Address): boolean => {
...
   if (asset) {
-    const assetAddress = asset.symbol.slice(asset.ticker.length + 1)
+    const assetAddress = asset.symbol.slice(asset.ticker.length + 1).toLowerCase()
```
veado, line 95: In general it would be nice to create ETH/ERC20 `Asset`'s based on valid check sum addresses only. Then we would not need to use `toLowerCase()`...

packages/xchain-ethereum/src/utils.ts
```diff
@@ -90,15 +90,12 @@ export const validateAddress = (address: Address): boolean => {
...
  */
-export const getTokenAddress = (asset: Asset | null): string | null => {
-  if (asset) {
-    const assetAddress = asset.symbol.slice(asset.ticker.length + 1).toLowerCase()
-    if (validateAddress(assetAddress)) {
-      return assetAddress
-    }
+export const getTokenAddress = (asset: Asset): string | null => {
+  try {
+    return ethers.utils.getAddress(asset.symbol.slice(asset.ticker.length + 1).substr(2))
```
veado, line 95: Check first two characters and trim these if they are 0x|0X only.

packages/xchain-ethereum/src/utils.ts
```diff
@@ -92,7 +92,12 @@ export const validateAddress = (address: Address): boolean => {
  */
 export const getTokenAddress = (asset: Asset): string | null => {
   try {
-    return ethers.utils.getAddress(asset.symbol.slice(asset.ticker.length + 1).substr(2))
+    const assetAddress = asset.symbol.slice(asset.ticker.length + 1)
+
+    if (assetAddress.startsWith('0x') || assetAddress.startsWith('0X'))
+      return ethers.utils.getAddress(assetAddress.substr(2))
```
veado, line 98: // strip 0X only - 0x is still valid
const strip0X = assetAddress.replace(/\^0X/, '')
ethers.utils.getAddress(strip0X)

veado: Since this `0X` it' s a THORChain/Midgard related "issue", we should handle it in ASGDX, but not in `xchain-eth`. `strip0X` might be a helper in `asgardex-util` then.

veado (approved):

PR #276, opened 2021-03-14
fix: litecoin export `Utils.validateAddress` and bitcoincash `Utils.validateAddress`

davidp94:
chore: fix mainnet urls for litecoin and bch


veado (approved):
LGTM

PR #292, opened 2021-04-08
Add `feeOptionKey` for `approve`

junkai121:


packages/xchain-ethereum/src/client.ts
```diff
@@ -601,16 +606,35 @@ export default class Client implements XChainClient, EthereumClient {
...
+            .toFixed(),
+        )
+      const gasLimit = await this.estimateApprove(spender, sender, amount).catch(() => undefined)
+
       const txAmount = amount ? BigNumber.from(amount.amount().toFixed()) : MAX_APPROVAL
       const txResult = await this.call<TransactionResponse>(sender, erc20ABI, 'approve', [
         spender,
         txAmount,
-        { from: this.getAddress() },
+        { from: this.getAddress(), gasPrice: gasPrice, gasLimit },
```
veado, line 637: [nitpick] `gasPrice: gasPrice` -> `gasPrice`

packages/xchain-ethereum/src/client.ts
```diff
@@ -78,7 +78,12 @@ export interface EthereumClient {
   estimateFeesWithGasPricesAndLimits(params: FeesParams): Promise<FeesWithGasPricesAndLimits>
 
   isApproved(spender: Address, sender: Address, amount: BaseAmount): Promise<boolean>
-  approve(spender: Address, sender: Address, amount?: BaseAmount): Promise<TransactionResponse>
+  approve(
+    spender: Address,
+    sender: Address,
+    feeOptionKey?: FeeOptionKey,
+    amount?: BaseAmount,
+  ): Promise<TransactionResponse>
```
veado, line 86: [nitpick] Use an object as parameter

packages/xchain-ethereum/src/client.ts
```diff
@@ -601,16 +606,35 @@ export default class Client implements XChainClient, EthereumClient {
...
+        feeOptionKey &&
+        BigNumber.from(
+          (
+            await this.estimateGasPrices()
+              .then((prices) => prices[feeOptionKey])
+              .catch(() => getDefaultGasPrices()[feeOptionKey])
+          )
+            .amount()
+            .toFixed(),
+        )
```
veado, line 630: I would avoid such nested things ^, since you can't know the result in case of errors. Even if you wrap everything into a `try/catch` statement before, the error message might be messed up and it might be hard to understand). Also, the code is not well readable anymore at some point. Just my 2 cents.

packages/xchain-ethereum/src/client.ts
```diff
@@ -79,10 +80,9 @@ export interface EthereumClient {
 
   isApproved(spender: Address, sender: Address, amount: BaseAmount): Promise<boolean>
   approve(
-    spender: Address,
-    sender: Address,
-    feeOptionKey?: FeeOptionKey,
-    amount?: BaseAmount,
+    params: ApproveParams & {
+      feeOptionKey?: FeeOptionKey
+    },
```
veado, line 85: Let's add `feeOptionKey` to `ApproveParams`. And for `estimateApprove` use `Omit<ApproveParams, 'feeOptionKey'>` as parameter.

veado (changes requested):

veado (approved):

PR #298, opened 2021-04-13
Add util helpers for explorer urls

junkai121:


veado (commented):
Can we use new `utils` helpers in `client`, too (to avoid duplication of same code)?

junkai121:
> Can we use new `utils` helpers in `client`, too (to avoid duplication of same code)?

In client side, it uses the explorer urls from constructor. Currently user can set explorer url when creating client.
BTW in utils helper, it will use the default explorer urls.

packages/xchain-thorchain/src/util.ts
```diff
@@ -241,3 +241,58 @@ export const getDefaultFees = (): Fees => {
...
+  }
+}
+
+/**
+ * Get the explorer url.
+ *
+ * @param {Network} network
+ * @returns {string} The explorer url for thorchain based on the given network.
+ */
+export const getExplorerUrl = (network: Network): string => {
```
veado, line 263: `getExplorerUrl` -> `getDefaultExplorerUrlByNetwork`

packages/xchain-thorchain/src/util.ts
```diff
@@ -241,3 +241,58 @@ export const getDefaultFees = (): Fees => {
...
+}
+
+/**
+ * Get the explorer url for the given address.
+ *
+ * @param {Network} network
+ * @param {Address} address
+ * @returns {string} The explorer url for the given address.
+ */
+export const getExplorerAddressUrl = (network: Network, address: Address): string => {
```
veado, line 274: `getExplorerAddressUrl` -> `getDefaultAddressUrl`

packages/xchain-thorchain/src/util.ts
```diff
@@ -241,3 +241,58 @@ export const getDefaultFees = (): Fees => {
...
+}
+
+/**
+ * Get the explorer url for the given node.
+ *
+ * @param {Network} network
+ * @param {Address} node address
+ * @returns {string} The explorer url for the given node.
+ */
+export const getExplorerNodeUrl = (network: Network, address: Address): string => {
```
veado, line 285: `getExplorerNodeUrl` -> `getDefaultExplorerNodeUrl`

veado (approved):

PR #301, opened 2021-04-16
[xchain-thorchain] Change gas.fee to `auto` in deposit

stuartkuentzel:
as a temporary fix, the value returned buildDepositTx for fee.gas is causing errors in the "deposit" function, where users with sufficient funds are being rejected with error "insufficient funds".
hardcodes gas.fee for this check to ~`1000000`~ `auto`.

Since we have seen it might still fail for first attempt, we try to send deposit tx up to three times.


packages/xchain-thorchain/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# v.0.13.3-1 (2021-04-16)
```
veado, line 1: `v.0.13.4`

packages/xchain-thorchain/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@xchainjs/xchain-thorchain",
-  "version": "0.13.3",
+  "version": "0.13.3-1",
```
veado, line 3: `v.0.13.4`

packages/xchain-thorchain/src/client.ts
```diff
@@ -582,6 +582,7 @@ class Client implements ThorchainClient, XChainClient {
       })
       const unsignedStdTx = await this.buildDepositTx(msgNativeTx)
       const fee = unsignedStdTx.fee
+      fee.gas = '1000000' // fee.gas temporarily hardcoded since the value returned is too high, causing "insufficient funds" errors for users with enough RUNE
```
veado, line 585: Not sure if `1000000` is correct, we discussed `10000000000` internally

cc/ @cbarraford

veado: We have decided to set `gas = 'auto'`. Since we have seen it might still fail for first attempt, we try to send `deposit` tx up to three times.

PR #303, opened 2021-04-19
[xchain-thorchain] Change fee.gas to `1000000` in deposit

stuartwk:
updates fee.gas to 1000000 (one million) in deposit


veado (approved):

skyorion427 (approved):
Tested send/deposit and works well.

LGTM!

PR #305, opened 2021-04-19
feat: export BTC.`calcFee`

liammcaweeney:


.idea/xchainjs-lib.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
davidp94: @liammcaweeney not sure we need this

davidp94 (changes requested):
Need to add changelog and bump the version in package.json

davidp94 (approved):

Issue #233, opened 2021-02-18
ADD: HD Wallet


thorchain-admin:
Allow requesting HD Wallet

```
getAddress(index)
```

Index is optional, default 0, if set, the return the address for the index at that level. 

- [x] Bitcoin
- [x] Ethereum
- [x] Binance
- [x] THORChain
- [x] Litecoin
- [x] BitcoinCash

davidp94:
I think we should add it as the instantiation of the `client` optionally as we'd need to sign depending of the index too @thorchain-admin @junkai121

mikewyszinski:
So i've picked up this card and have a few questions, remember i'm new to this project so perhaps I don't have a full picture yet.

1. the original request mentions only getAddress(index), however the current branch also changes the Client constructor to 
```
rootPath?:string
index?:number
```
I'm inclined to back this change out for 2 reasons:

    1. It adds more state to the client libraries when i feel like the less state we manage, the better.
         An example of the tipple effect is that setNetwork() now has to manage [resetting these during a network change(https://github.com/xchainjs/xchainjs-lib/blob/bdcc50346395e9609ee208785e86de20079afeb9/packages/xchain-bitcoin/src/client.ts#L141)
     2.  do we really want consumers of the library need to know the derivations paths or override the derivation paths? I'm inclined to move ALL root paths (including the ledger specific paths )  as constants/ or default configs

2.  seems like setPhrase() returns an address as a convenience, recommend removing this as now we would either needto add an index OR default to index =0 , both of which seem unintuitive for a setter.

davidp94:
> [](https://github.com/xchainjs/xchainjs-lib/blob/bdcc50346395e9609ee208785e86de20079afeb9/packages/xchain-bitcoin/src/client.ts#L141)
> 2. do we really want consumers of the library need to know the derivations paths or override the derivation paths? I'm inclined to move ALL root paths (including the ledger specific paths ) as constants/ or default configs

@veado @thorchain-admin

mikewyszinski:
OK pushing an interim version ...

#  status

-  Bitcoin coded and unit tests passing
-  Litecoin coded and unit tests passing
[...]

Issue #13, opened 2020-10-19
ADD: BIP44 Spec address derivation


thorchain-admin:
# Current

Phrase is used directly for entropy

# Desired

BIP 44 compatibility with other wallets. However the client will only use Index 0, but can in future handle multiple indexes at the same time (eg, like using "sub-accounts" in a single wallet). 

## Hardcode SLIP44 path

Hard code in the SLIP44 path into each client as a private variable
https://github.com/satoshilabs/slips/blob/master/slip-0044.md

## Derive key && address using Index 0
`getAddress(): Address`

This should derive using BIP44, but index 0 only. 

# Implementation 

- [x] Bitcoin (`84/0/0/0`)

- [x] Binance Chain  (`44/714/0/0`)

- [x] Ethereum  (`44/30/0/0`)

- [x] THORChain  (`44/931/0/0`)

- [x] Cosmos ( `44/118/0/0`)

Thorian1te:
Closing this issue. nothing heard in a while

PR #308, opened 2021-04-24
feat: HD `derivationPath`

davidp94:
Implement #233 

- Added `derivationPath` as an optional parameter to constructor of the clients below.
- Added `derivationPath` as an optional parameter to `setPhrase`

- [x] ethereum  -- example of derivationPath `m/44'/60'/0'/0/5`
- [x] bitcoin -- example of derivationPath `m/84'/0'/0'/0/1`
- [x] bitcoincash -- example of derivationPath `m/44'/145'/0'/0/1`
- [x] litecoin -- example of derivationPath `m/84'/2'/0'/0/1`
- [x] thorchain -- example of derivationPath `44'/931'/0'/0/1`
- [x] binance -- example of derivationPath `2` (this is just the stringified index as the SDK doesn't allow to specify the whole path)

@mikewyszinski notes:

```
    some derivation paths for mainnet/testnet are currently defaulted identical (ex. cosmos), this was to allow the tests to pass as written, to ensure backwards compatibilty with any wallets
    since most of the state has been removed to make managing derivation paths simpler, we added more time so the tests execute more slowly vs. master
    ethereum - master - 80 secs
    ethereum - hd wallet - 180 secs
    there are still parts on the implementations that have the old implementation commented out , if this PR is accepted, i can walk through the code and make sure we delete these
    if the plan is to on-board the top 50-100 coins & support HW wallets in this library, we should consider refactoring the interface to be more "pluggable". This would allow us to build/integrate chains/wallets/explorer/etc as "providers" individually This would make scaling out the lib easier and much easier to maintain.
```

### DIFF
--- a/packages/xchain-bitcoincash/src/client.ts
+++ b/packages/xchain-bitcoincash/src/client.ts
@@ -2,6 +2,7 @@ const bitcash = require('@psf/bitcoincashjs-lib')
 
 import * as utils from './utils'
 import {
+  RootDerivationPaths,
   Address,
   Balance,
   Network,
@@ -38,10 +39,6 @@ type BitcoinCashClientParams = XChainClientParams & {
   rootPath?: string
   index?: number
 }
-const rootPaths = {
-  mainnet: `m/44'/145'/0'/0/`,
-  testnet: `m/44'/1'/0'/0/`,
-}
 
 /**
  * Custom Bitcoin Cash client
@@ -52,6 +49,7 @@ class Client implements BitcoinCashClient, XChainClient {
   private haskoinUrl: ClientUrl
   private nodeUrl: ClientUrl
   private nodeAuth?: NodeAuth
+  private rootDerivationPaths: RootDerivationPaths
 
   /**
    * Constructor
@@ -74,10 +72,15 @@ class Client implements BitcoinCashClient, XChainClient {
       username: 'thorchain',
       password: 'password',
     },
+    rootDerivationPaths = {
+      mainnet: `m/44'/145'/0'/0/`,
+      testnet: `m/44'/1'/0'/0/`,
+    },
   }: BitcoinCashClientParams) {
     this.network = network
     this.haskoinUrl = haskoinUrl
     this.nodeUrl = nodeUrl
+    this.rootDerivationPaths = rootDerivationPaths
     phrase && this.setPhrase(phrase)
     this.nodeAuth =
       // Leave possibility to send requests without auth info for user
@@ -258,7 +261,7 @@ class Client implements BitcoinCashClient, XChainClient {
   getAddress = (index = 0): Address => {
     if (this.phrase) {
       try {
-        const derivationPath = rootPaths[this.network] + `${index}`
+        const derivationPath = this.rootDerivationPaths[this.network] + `${index}`
         const keys = this.getBCHKeys(this.phrase, derivationPath)
         const address = keys.getAddress(index)
 
@@ -289,8 +292,9 @@ class Client implements BitcoinCashClient, XChainClient {
    *
    * @throws {"Invalid address"} Thrown if the given address is an invalid address.
    */
-  getBalance = async (address?: string): Promise<Balance[]> => {
-    return utils.getBalance({ haskoinUrl: this.getHaskoinURL(), address: address || this.getAddress() })
+  getBalance = async (index = 0): Promise<Balance[]> => {
+    const address = this.getAddress(index)
+    return utils.getBalance({ haskoinUrl: this.getHaskoinURL(), address })
   }
 
   /**
@@ -304,12 +308,16 @@ class Client implements BitcoinCashClient, XChainClient {
    */
   getTransactions = async ({ address, offset, limit }: TxHistoryParams): Promise<TxsPage> => {
     try {
-      address = address || this.getAddress()
+      const theAddress = typeof address === 'number' ? this.getAddress(address) : address + ''
       offset = offset || 0
       limit = limit || 10
 
-      const account = await getAccount({ haskoinUrl: this.getHaskoinURL(), address })
-      const txs = await getTransactions({ haskoinUrl: this.getHaskoinURL(), address, params: { offset, limit } })
+      const account = await getAccount({ haskoinUrl: this.getHaskoinURL(), address: theAddress })
+      const txs = await getTransactions({
+        haskoinUrl: this.getHaskoinURL(),
+        address: theAddress,
+        params: { offset, limit },
+      })
 
       if (!account || !txs) {
         throw new Error('Invalid address')
@@ -423,8 +431,8 @@ class Client implements BitcoinCashClient, XChainClient {
    */
   transfer = async (params: TxParams & { feeRate?: FeeRate }): Promise<TxHash> => {
     try {
-      const index = 0 //shoudl we be using index 0 here?
-      const derivationPath = rootPaths[this.network] + `${index}`
+      const index = params.from || 0
+      const derivationPath = this.rootDerivationPaths[this.network] + `${index}`
 
       const feeRate = params.feeRate || (await this.getFeeRates()).fast
       const { builder, utxos } = await utils.buildTx({
--- a/packages/xchain-bitcoincash/src/client.ts
+++ b/packages/xchain-bitcoincash/src/client.ts
@@ -35,6 +35,8 @@ type BitcoinCashClientParams = XChainClientParams & {
   haskoinUrl?: ClientUrl
   nodeUrl?: ClientUrl
   nodeAuth?: NodeAuth
+  rootPath?: string
+  index?: number
 }
 
 /**
@@ -43,7 +45,9 @@ type BitcoinCashClientParams = XChainClientParams & {
 class Client implements BitcoinCashClient, XChainClient {
   private network: Network
   private phrase = ''
-  private derivationPath: string =''
+  private rootPath = ''
+  private index = 0
+  private derivationPath = ''
   private haskoinUrl: ClientUrl
   private nodeUrl: ClientUrl
   private nodeAuth?: NodeAuth
@@ -69,17 +73,21 @@ class Client implements BitcoinCashClient, XChainClient {
       username: 'thorchain',
       password: 'password',
     },
-    derivationPath,
+    rootPath,
+    index = 0,
   }: BitcoinCashClientParams) {
     this.network = network
     this.haskoinUrl = haskoinUrl
     this.nodeUrl = nodeUrl
-    this.derivationPath = derivationPath || this.derivePath()
+    this.rootPath = rootPath || utils.getRootPath(network)
+    this.index = index
+    this.derivationPath = this.derivePath()
+    phrase && this.setPhrase(phrase, index)
     this.nodeAuth =
       // Leave possibility to send requests without auth info for user
       // by strictly passing nodeAuth as null value
       nodeAuth === null ? undefined : nodeAuth
-    phrase && this.setPhrase(phrase, this.derivationPath)
+    phrase && this.setPhrase(phrase, index)
   }
 
   /**
@@ -130,10 +138,11 @@ class Client implements BitcoinCashClient, XChainClient {
    * @throws {"Invalid phrase"}
    * Thrown if the given phase is invalid.
    */
-  setPhrase = (phrase: string, derivationPath?: string): Address => {
+  setPhrase = (phrase: string, index = 0): Address => {
     if (validatePhrase(phrase)) {
       this.phrase = phrase
-      this.derivationPath = derivationPath || this.derivePath()
+      this.index = index
+      this.derivationPath = this.derivePath()
       const address = this.getAddress()
       return address
     } else {
@@ -159,11 +168,13 @@ class Client implements BitcoinCashClient, XChainClient {
    * @throws {"Network must be provided"}
    * Thrown if network has not been set before.
    */
-  setNetwork = (network: Network): void => {
-    if (network) {
-      this.network = network
-    } else {
+  setNetwork = (net: Network): void => {
+    if (!net) {
       throw new Error('Network must be provided')
+    } else {
+      this.network = net
+      this.rootPath = utils.getRootPath(net)
+      this.derivationPath = this.derivePath()
     }
   }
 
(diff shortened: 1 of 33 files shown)


davidp94:
To reviewers:


Let me know where I should add more comments; documentation; and what other tests cases I should look at; (maybe you can give me your own derivation paths and results from other sources than here!)

For thorchain I am not sure if the addresses path /1 are good; please double check.

liammcaweeney (approved):
LGTM

veado:
@davidp94 Could you give some more information about your goal with this PR, pls? Do you need to handle different `pathes` or just different `indexes`? Currently `xchain-*` libs have paths hard-coded as defined in #13 to handle different `indexes` (but not `paths`) in the future. Check following note:

> BIP 44 compatibility with other wallets. However the client will only use Index 0, but can in future handle multiple indexes at the same time (eg, like using "sub-accounts" in a single wallet).

^ https://github.com/xchainjs/xchainjs-lib/issues/13

davidp94:
> @davidp94 Could you give some more information about your goal with this PR, pls? Do you need to handle different `paths` or just different `indexes`? Currently `xchain-*` libs have paths hard-coded as defined in #13 to handle different `indexes` (but not `paths`) in the future. Check following note:
> 
> > BIP 44 compatibility with other wallets. However the client will only use Index 0, but can in future handle multiple indexes at the same time (eg, like using "sub-accounts" in a single wallet).
> 
> ^ #13

Goal is to handle different paths (which includes indexes supports); not only indexes.

The fact is that Ledger for ethereum used different paths in legacy version so this approach allows us to be more flexible (cf. https://www.reddit.com/r/ledgerwallet/comments/bpoyoa/which_derivation_path_m44600_or_m4460/ )

Actually #13 might be included for this PR too.

packages/xchain-bitcoincash/src/client.ts
```diff
@@ -289,8 +292,9 @@ class Client implements BitcoinCashClient, XChainClient {
    *
    * @throws {"Invalid address"} Thrown if the given address is an invalid address.
    */
-  getBalance = async (address?: string): Promise<Balance[]> => {
-    return utils.getBalance({ haskoinUrl: this.getHaskoinURL(), address: address || this.getAddress() })
+  getBalance = async (index = 0): Promise<Balance[]> => {
```
davidp94, line 295: Hmm I don't think we should do that for `getBalance`

we should keep `getBalance(address?: string` imo

cc. @veado

davidp94:
Ha yes right; maybe we should still keep getAddress ({address?: string}) ?
if we want to get the balance for another one.

or we would give const secondAddr = getAddress(1)
then getAddress(secondAddr)

packages/xchain-cosmos/src/client.ts
```diff
@@ -48,9 +51,17 @@ class Client implements CosmosClient, XChainClient {
    *
    * @throws {"Invalid phrase"} Thrown if the given phase is invalid.
    */
-  constructor({ network = 'testnet', phrase }: XChainClientParams) {
+  constructor({
+    network = 'testnet',
+    phrase,
+    rootDerivationPaths = {
+      mainnet: `44'/118'/0'/0/`,
+      testnet: `44'/118'/0'/0/`, //TODO: shoudn't this be 44'/118'/0'/1/ ?
```
davidp94, line 59: Hmm. not sure.

cc, @veado

packages/xchain-cosmos/src/client.ts
```diff
@@ -62,8 +73,8 @@ class Client implements CosmosClient, XChainClient {
    */
   purgeClient(): void {
     this.phrase = ''
-    this.address = ''
-    this.privateKey = null
+    // this.address = ''
```
davidp94, line 76: Need to remove;

packages/xchain-cosmos/src/client.ts
```diff
@@ -203,27 +219,24 @@ class Client implements CosmosClient, XChainClient {
...
-
   /**
    * Validate the given address.
    *
    * @param {Address} address
    * @returns {boolean} `true` or `false`
    */
   validateAddress = (address: Address): boolean => {
-    return this.sdkClient.checkAddress(address)
+    return this.getCosmosClientSdk(0).checkAddress(address)
```
davidp94, line 239: This makes performance pretty slow isn't it?

cc @veado

packages/xchain-cosmos/src/client.ts
```diff
@@ -282,9 +294,9 @@ class Client implements CosmosClient, XChainClient {
       this.registerCodecs()
 
       const mainAsset = this.getMainAsset()
-      const txHistory = await this.sdkClient.searchTx({
+      const txHistory = await this.getCosmosClientSdk(0).searchTx({
```
davidp94, line 297: Again reinstantiating this is heavy; should we have a "anonClient" instantiated used for this kind of function call?

packages/xchain-cosmos/src/client.ts
```diff
@@ -308,7 +320,7 @@ class Client implements CosmosClient, XChainClient {
    */
   getTransactionData = async (txId: string): Promise<Tx> => {
     try {
-      const txResult = await this.sdkClient.txsHashGet(txId)
+      const txResult = await this.getCosmosClientSdk(0).txsHashGet(txId)
```
davidp94, line 323: Same as previously

packages/xchain-ethereum/__tests__/client.mainnet.test.ts
```diff
@@ -48,67 +64,67 @@ describe('Client Test', () => {
...
+  //       price: {
+  //         rate: 1196.5425814145788,
+  //         diff: 11.71,
+  //         diff7d: 62.3,
+  //         ts: 1609987982,
+  //         marketCapUsd: 136582198332.62915,
+  //         availableSupply: 114147377.999,
+  //         volume24h: 44933107598.39366,
+  //         diff30d: 108.688017487141,
+  //         volDiff1: 7.230942506781318,
```
davidp94, line 95: why is it commented out?

packages/xchain-ethereum/__tests__/client.testnet.test.ts
```diff
@@ -200,11 +201,9 @@ describe('Client Test', () => {
     expect(balance[1].amount.amount().isEqualTo(baseAmount('96713467036431546', 18).amount())).toBeTruthy()
   })
 
-  it('throws error on bad address', async () => {
```
davidp94, line 203: Need to keep that test. `getBalance('badAddress')`

packages/xchain-ethereum/__tests__/client.testnet.test.ts
```diff
@@ -529,13 +528,15 @@ describe('Client Test', () => {
     )
 
     let isApproved = await ethClient.isApproved(
+      0,
```
davidp94, line 531: Why don't we pass the address here?

packages/xchain-ethereum/__tests__/client.testnet.test.ts
```diff
@@ -529,13 +528,15 @@ describe('Client Test', () => {
...
       '0x8c2a90d36ec9f745c9b28b588cba5e2a978a1656',
       '0x4f96fe3b7a6cf9725f59d353f723c1bdb64ca6aa',
       baseAmount(100, ETH_DECIMAL),
     )
     expect(isApproved).toEqual(true)
 
     isApproved = await ethClient.isApproved(
+      0,
```
davidp94, line 539: same as above

packages/xchain-ethereum/__tests__/client.testnet.test.ts
```diff
@@ -555,6 +556,7 @@ describe('Client Test', () => {
     mock_all_api(etherscanUrl, ropstenInfuraUrl, ropstenAlchemyUrl, 'eth_estimateGas', '0x5208')
 
     const gasLimit = await ethClient.estimateApprove({
+      walletIndex: 0,
```
davidp94, line 559: I think we can put address: string as a parameter as this function is not signing;

packages/xchain-ethereum/__tests__/client.testnet.test.ts
```diff
@@ -581,6 +583,7 @@ describe('Client Test', () => {
     )
 
     const tx = await ethClient.approve({
+      walletIndex: 0,
```
davidp94, line 586: This seems OK;

packages/xchain-ethereum/__tests__/client.testnet.test.ts
```diff
@@ -600,13 +603,19 @@ describe('Client Test', () => {
...
 
-    const gasLimit = await ethClient.estimateCall('0xd15ffaef3112460bf3bcd81087fcbbce394e2ae7', erc20ABI, 'transfer', [
-      '0x8c2a90d36ec9f745c9b28b588cba5e2a978a1656',
-      BigNumber.from(baseAmount('10000000000000', ETH_DECIMAL).amount().toString()),
-      {
-        from: ethClient.getAddress(),
-      },
-    ])
+    const gasLimit = await ethClient.estimateCall(
+      0,
```
davidp94, line 607: Should be the address

packages/xchain-ethereum/src/client.ts
```diff
@@ -66,8 +67,15 @@ import {
  */
 export interface EthereumClient {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  call<T>(asset: Address, abi: ethers.ContractInterface, func: string, params: Array<any>): Promise<T>
+  call<T>(
+    walletIndex: number,
```
davidp94, line 71: why is that needed?

packages/xchain-ethereum/src/client.ts
```diff
@@ -78,7 +86,7 @@ export interface EthereumClient {
   estimateGasLimit(params: FeesParams): Promise<BigNumber>
   estimateFeesWithGasPricesAndLimits(params: FeesParams): Promise<FeesWithGasPricesAndLimits>
 
-  isApproved(spender: Address, sender: Address, amount: BaseAmount): Promise<boolean>
+  isApproved(walletIndex: number, spender: Address, sender: Address, amount: BaseAmount): Promise<boolean>
```
davidp94, line 89: I don't think we need walletIndex for that call

packages/xchain-ethereum/src/client.ts
```diff
@@ -118,36 +124,22 @@ export default class Client implements XChainClient, EthereumClient {
     ethplorerUrl = 'https://api.ethplorer.io',
     ethplorerApiKey = 'freekey',
     explorerUrl,
-    phrase,
+    phrase = '',
```
davidp94, line 127: again; what happen if phrase = ''?

packages/xchain-ethereum/src/client.ts
```diff
@@ -118,36 +124,22 @@ export default class Client implements XChainClient, EthereumClient {
...
+    rootDerivationPaths = {
+      mainnet: `m/44'/60'/0'/0/`,
+      testnet: `m/44'/60'/0'/0/`, // this is INCORRECT but makes the unit tests pass
```
davidp94, line 130: Why is this incorrect? mainnet and testnet derivation paths are the same for ETH mainnet or testnet. 

https://iancoleman.io/bip39/

packages/xchain-ethereum/src/client.ts
```diff
@@ -187,11 +178,14 @@ export default class Client implements XChainClient, EthereumClient {
...
-    if (!this.address) {
+  getAddress = (index = 0): Address => {
+    if (index < 0) {
+      throw new Error('index must be greater than zero')
+    }
+    if (!this.phrase) {
       throw new Error('Phrase must be provided')
     }
-    return this.address
+    return this.getWallet(index).address.toLowerCase()
```
davidp94, line 188: Any particular reason of removing the checksum casing of the address?

cf. https://coincodex.com/article/2078/ethereum-address-checksum-explained/

packages/xchain-ethereum/src/client.ts
```diff
@@ -570,16 +563,17 @@ export default class Client implements XChainClient, EthereumClient {
    * Thrown if the given contract address is empty.
    */
   estimateCall = async (
-    address: Address,
+    walletIndex = 0,
```
davidp94, line 566: For an estimate call; we don't need to have the wallet seed phrase

We can just do

```
contract.estimateGas[func](...params)
```

And add `from: senderAddress` 

```
ct for a read-only method may include any of:

    overrides.from - the msg.sender (or CALLER) to use during the execution of the code
    overrides.value - the msg.value (or CALLVALUE) to use during the exectuiont of the code
    overrides.gasPrice - the price to pay per gas (theoretically); since there is no transaction, there is not going to be any fee charged, but the EVM still requires a value to report to tx.gasprice (or GASPRICE); most developers will not require this
    overrides.gasLimit - the amount of gas (theoretically) to allow a node to use during the execution of the code; since there is no transaction, there is not going to be any fee charged, but the EVM still processes gas metering so calls like gasleft (or GAS) report meaningful values
    overrides.blockTag - a block tag to simulate the execution at, which can be used for hypothetical historic analysis; note that many backends do not support this, or may require paid plans to access as the node database storage and processing requirements are much higher 
    ```

cf. https://docs.ethers.io/v5/api/contract/contract/

File: packages/xchain-ethereum/src/client.ts [hunk 20]
```diff
@@ -646,10 +650,15 @@ export default class Client implements XChainClient, EthereumClient {
    * @param {BaseAmount} amount The amount of token. By default, it will be unlimited token allowance. (optional)
    * @returns {BigNumber} The estimated gas limit.
    */
-  estimateApprove = async ({ spender, sender, amount }: Omit<ApproveParams, 'feeOptionKey'>): Promise<BigNumber> => {
+  estimateApprove = async ({
+    walletIndex = 0,
```
davidp94, line 654: estimation does not need that; same as `isApproved` and `estimateCall`

packages/xchain-thorchain/src/types/client-types.ts
```diff
@@ -21,6 +21,7 @@ export type ThorchainClientParams = {
 }
 
 export type DepositParam = {
+  from?: number
```
davidp94, line 24: walletIndex would be better

davidp94 (commented):
I think there's many area where we have `walletIndex` set up; especially non signing operations.

Need to make naming consistent.

Perforamnce wise; regenerating a wallet for some query calls is not a good thing

PR #313, opened 2021-05-06
Fix wrong eth balance issue due to internal txn

thorswapdev:
## Issue
https://github.com/EverexIO/Ethplorer/issues/209

## Changelog

- get ETH Balance directly using provider
- update `getBalance` method


veado (changes requested):
Code LGTM.

Could you check why tests not running on CI?

Pls update CHANGELOG + change version in package.json to be ready to publish next version

PR #314, opened 2021-05-06
Fix/eth balance

thorswapdev:
## Issue
https://github.com/EverexIO/Ethplorer/issues/209

## Changelog

- get ETH Balance directly using provider
- update `getBalance` method

### DIFF
--- a/packages/xchain-ethereum/CHANGELOG.md
+++ b/packages/xchain-ethereum/CHANGELOG.md
@@ -1,5 +1,11 @@
 # v.x.x.x
 
+# v.0.20.0 (2021-05-06)
+
+### Fix
+
+- Get ETH balance directly from provider
+  
 # v.0.19.0 (2021-05-05)
 
 ### Breaking change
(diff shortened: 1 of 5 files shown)


veado:
@thorswapdev Ping ^

thorswapdev:
> @thorswapdev Ping ^

@veado , I don't think it worth covering a unit test for eth getbalance since it's pretty much standard function from ethers.js lib. We can just remove the unit test that was relying on the ethplorer api response. 

Please let me know your thoughts. :)

veado:
> > @thorswapdev Ping ^
> 
> @veado , I don't think it worth covering a unit test for eth getbalance since it's pretty much standard function from ethers.js lib. We can just remove the unit test that was relying on the ethplorer api response.
> 
> Please let me know your thoughts. :)

@thorswapdev Agree, we don't want to test `provider.getBalances`. However, failing tests are about `getBalance` method of `Client` and as you know it includes more than just calling `provider.getBalances`. Removing such tests is really a bad idea - pls fix it.

PR #315, opened 2021-05-07
feat: support derivation path(binance)

smart-dev30:
support derivation path(sub account index) to binance client

### DIFF
--- a/packages/xchain-binance/src/client.ts
+++ b/packages/xchain-binance/src/client.ts
@@ -35,7 +35,7 @@ import {
   BNBChain,
   assetToString,
 } from '@xchainjs/xchain-util'
-import * as xchainCrypto from '@xchainjs/xchain-crypto'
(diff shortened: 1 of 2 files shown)


packages/xchain-binance/src/client.ts
```diff
@@ -247,18 +222,15 @@ class Client implements BinanceClient, XChainClient {
...
-      if (!address) {
-        throw new Error(
-          'Address has to be set. Or set a phrase by calling `setPhrase` before to use an address of an imported key.',
-        )
-      }
-
-      this.address = address
+  getAddress = (index = 0): string => {
+    const address = crypto.getAddressFromPrivateKey(this.getPrivateKey(index), getPrefix(this.network))
+    if (!address) {
```
mikewyszinski, line 227: do we need this check? it doesn't hurt but may not be neccessary..

smart-dev30: nice catch! updated with 4b76ae0

PR #326, opened 2021-05-24
Fix bitcoincash transaction prefixed address

thatStrangeGuyThorchain:
- `utils:parseTransaction` added stripping prefixes to the address mapper
- bump version to `v0.10.1`

### DIFF
--- a/packages/xchain-bitcoincash/src/utils.ts
+++ b/packages/xchain-bitcoincash/src/utils.ts
@@ -14,6 +14,8 @@ import {
   GetChangeParams,
   UTXOs,
   UTXO,
+  TransactionInput,
+  TransactionOutput,
 } from './types'
 import { getAccount, getRawTransaction, getUnspentTransactions } from './haskoin-api'
 import { Network as BCHNetwork, TransactionBuilder } from './types/bitcoincashjs-types'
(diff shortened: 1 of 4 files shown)


stuartwk (approved):

packages/xchain-bitcoincash/src/utils.ts
```diff
@@ -192,7 +192,7 @@ export const parseTransaction = (tx: Transaction): Tx => {
       .map(
         (input) =>
           ({
-            from: input.address,
+            from: input.address ? stripPrefix(input.address) : input.address,
```
veado, line 195: Sounds confusing ^. Add a comment to understand `input.address` will always be available, because `input` with available `address` will use here only.

packages/xchain-bitcoincash/src/utils.ts
```diff
@@ -201,7 +201,7 @@ export const parseTransaction = (tx: Transaction): Tx => {
       .map(
         (output) =>
           ({
-            to: output.address,
+            to: output.address ? stripPrefix(output.address) : output.address,
```
veado, line 204: As same as previous comment

veado (changes requested):

veado (approved):

PR #329, opened 2021-05-27
Fix eth balance issue

thorswapdev:
- Get ETH balance directly from the provider
- Remove unit test for `getBalance` method
- Bump up version to `0.21.0`

### DIFF
--- a/packages/xchain-ethereum/CHANGELOG.md
+++ b/packages/xchain-ethereum/CHANGELOG.md
@@ -1,3 +1,9 @@
+# v.0.21.0 (2021-05-27)
+
+### Fix
+
+- Get ETH balance directly from provider
+
 # v.0.20.0 (2021-05-17)
 
 ### Breaking change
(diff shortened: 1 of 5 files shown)


veado (changes requested):
@thorchain-admin [As discussed before](https://github.com/xchainjs/xchainjs-lib/pull/314#issuecomment-848458032) don't delete unit tests for `getBalance`, but update tests to cover latest change. Again, it's not about to test `getBalances` of any provider, it's about to test parsing / transformation of balances in `Client.getBalances` into `Balances` to check it still works as expected.

thorswapdev:
getBalance method is using eth provider for getting eth balance, and therefore unit test won't work out.
@veado , if you think unit test is mandatory, we should start writing the unit test for all methods, not only for getBalance and make sure 100% of test coverage.
If you can't approve the PR because unit test isn't covered, I would complete this PR later, not now.

thorswapdev:
https://github.com/xchainjs/xchainjs-lib/pull/329/files#diff-defd1dd34d821704121a28974af9cf3ca8bc89cb22cd7296be6e0d6db2c4583bR343

`client/getBalance` method is calling calling `provider.getBalance` inside and any unit test calling the getBalance method will be failing.

The previous unit test was only based on the REST API with mock response so it was possible but we can't cover the test for this function.

Closing this PR for now.

PR #333, opened 2021-05-30
export client params types which are used in constructor

severinwullschleger:
The ClientParams types which are used in the constructor of the client are now also exported. Then I can use the type also when I'm initializing the clients.


veado (approved):
LGTM

Thanks!

PR #337, opened 2021-05-31
Fix dust error occurred by accumulating memo output

thorswapdev:
### @xchainjs/xchain-bitcoin

- [x] Remove memo output from `accumulative`
- [x] Add memo output using `psbt.addOutput`

### @xchainjs/xchain-bitcoincash

- [x] Remove memo output from `accumulative`
- [x] Add memo output using `transactionBuilder`

### @xchainjs/xchain-litecoin

- [x] Remove memo output from `accumulative`
- [x] Add memo output using `psbt.addOutput`

### DIFF
--- a/packages/xchain-bitcoin/CHANGELOG.md
+++ b/packages/xchain-bitcoin/CHANGELOG.md
@@ -1,3 +1,9 @@
+# v.0.15.2 (2021-05-31)
+
+### Breaking change
+- don't add memo output to `coinselect/accumulative`
+- add memo output by using `psbt.addOutput` 
+
 # v.0.15.0 (2021-05-28)
 
 ### Breaking change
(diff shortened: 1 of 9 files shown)


mikewyszinski (approved):

mikewyszinski:
merging fix, but we may be introducing an edge case where memo's add to size of tx, but they will not be calculated in accumulate(), this leaves a possibility of submitting a tx without enough inputs to cover the fees added by the memo...

thorswapdev:
> merging fix, but we may be introducing an edge case where memo's add to size of tx, but they will not be calculated in accumulate(), this leaves a possibility of submitting a tx without enough inputs to cover the fees added by the memo...

that's definitely a possible but not ideally reasonable case.

1. If input can't cover < 40 bytes of `OP_RETURN` memo, it should be considered as dust. We are currently allowing only `>10k sats` amount for transactions to be captured by network.
2. fees are auto calculated based on the `input` and `output` by `psbt` transaction builder. Since `OP_RETURN` memo is added to the `psbt.output`, fee will be calculated correctly.

We will keep monitoring if edge case could happen but I guarantee it won't be happening in any case. :)

PR #338, opened 2021-06-01
refactored utils.buildTx() in BCH,LTC,BTC

### LINKED ISSUE #336: [BUG] dust error tx when accumulating memo output with 0 value
## Issue Description

when adding a memo with `0 value` to `outputs` using `coinselect/accumulative` , psbt composite doesn't work properly,
resulting `dust` error when broadcasting the raw tx.

## Solution

- don't add memo output using `coinselect/accumulative`.
- add memo using `psbt.addOutput` separately.

mikewyszinski:
this is a followup to fix #336, where we leave the memo in the ouputs to make sure accumulate includes enough inputs to cover the tx costs if a memo is included. we then re-add the memo using psbt.addOutput() to avoid the tx to be flagged with dust error.


thorswapdev (approved):
LGTM!

Issue #339, opened 2021-06-02
FIX: LTC adding memo twice and updating peer dependencies


mikewyszinski:
- xchain_litecoin - utils.buildTx() has leftover code from a refactor
- update peerDependencies
- reverting BCH utils.buildTx()

PR #340, opened 2021-06-02
fixes for #339

mikewyszinski:
fixes for #339


packages/xchain-bitcoincash/src/utils.ts
```diff
@@ -290,18 +285,23 @@ export const buildTx = async ({
...
-        const out = bitcash.address.toOutputScript(toLegacyAddress(output.address), bchNetwork(network))
-        transactionBuilder.addOutput(out, output.value)
+        out = bitcash.address.toOutputScript(toLegacyAddress(output.address), bchNetwork(network))
       }
+      transactionBuilder.addOutput(out, output.value)
     })
 
+    // add output for memo
+    if (compiledMemo) {
+      transactionBuilder.addOutput(compiledMemo, 0) // Add OP_RETURN {script, value}
```
stuartwk, line 302: is this adding the memo twice? I see on line 290 out is also being set as compiled memo?

thorswapdev (approved):
LGTM!

thorswapdev:
@mikewyszinski , pls fix conflicts and let's get it published.

packages/xchain-bitcoincash/src/utils.ts
mikewyszinski: thanks for catching this !

PR #361, opened 2021-06-19
More Promises In More Places

reidrankin:
I'm working on support for hardware wallets, and I've noticed that there are some breaking API changes that would be needed first:
- Since hardware wallets need to make asynchronous calls to get unlocked or pull their addresses, at least `setPhrase()` and `getAddress()` need to be asynchronous.
- Because constructors are always synchronous (and thus can't wait for a device to be unlocked), `new Client(params)` needs to become `await Client.create(params)`, with the constructor marked as `protected`.
- `purgeClient()` may not occur immediately on a hardware wallet, so either it needs to be asynchronous as well or a locking mechanism needs to be used to block any subsequent asynchronous operations from beginning until the device is in a consistent state. (This is far from impossible, but as long as we're considering breaking changes it seems a shame not to just go ahead and make it async.)

I have working test branches of `xchainjs-lib` and `asgardex-electron` with KeepKey support -- I even plumbed up React components for handling PIN entry! -- but the effort to merge all of it starts with getting consensus on this set of changes.

Personally, I'd love to see full, consistent use of Promises by all externally-exposed methods. This is because hardware wallets aren't the only cool thing you can do with Promises. A lot of newfangled APIs -- WebCrypto, Web Workers, WASM, and more -- are Promise-first or Promise-only. Imagine:
- Keys stored as non-exportable WebCrypto objects, for simultaneous security and native-like crypto performance
- A thin wrapper implementing `XChainClient` that forwards to an instance in a Web Worker, so expensive crypto operations don't block the UI
- Some sort of thing with Rust and WASM -- IDK exactly, but I bet it could be pretty cool

This PR is mergable as-is, but the commits are ordered by importance for ease of consideration.
- eed6957 is strictly necessary for hardware wallet support
- ea5e1f7 through 9eff341 complete the set of methods on the `XChainClient` interface
- 082030a through 6df6d16 cover the extra methods exposed by the chain-specific `FoobarClient` interface

One issue that might become obvious is that upgrading *all* this stuff adds a lot of extra `await`s. I agree. I do feel there's ample room for cleanup and re-org to remove most of the added messiness, but I wanted to present this PR as a set of clear, easy-to-review single-concern commits, and I felt that taking advantage of some of the opportunities for cleanup would make the diffs much harder to read. I intend to include some of that cleanup as part of a followup PR, which will focus on an internal, non-breaking refactor to improve maintainability.

One thing that could be done _right now_ to improve the readability of this changeset -- and maintainability of the project overall -- would be to limit what methods are exposed as part of the chain-specific client interfaces. For example, I have doubts as to how useful `xchain-polkadot`'s `getSS58Format()` and `getWsEndpoint()` methods are to a client, and just deleting those two lines from the `PolkadotClient` interface would allow f45f83a, 4db9682, and 0eb38c3 to all be dropped. Similar simplifications seem possible for Binance, Cosmos, and Thorchain as well.


reidrankin:
Upon re-reading this, I'm realizing that I haven't made it clear that my goal here isn't necessarily to merge all this, but to start a discussion about how the API should evolve and what breaking changes are acceptable in the pursuit of progress. The first commit in this list is necessary, but the rest are just one example of working out an approach 

I'm also preparing another PR for consideration which takes an alternative approach where each chain's client object is separated into a parameter object which contains the synchronous getters and setters and a delegate object which processes asynchronous actions. This is more invasive in terms of lines of code changed, but it allows a lot of repeated boilerplate to be factored out into a common prototype.

I'd hate for this to be merged without the opportunity to compare it against that approach, so I'm going to mark this as a "Draft".

veado:
@reidrankin Great thoughts, great work! How about to start changes for one `xchain-*` lib only by creating/using a new (async based) interface for `interface XChainClient`? It will minimize work for this first approach and will be easier to discuss because of less code changes.

reidrankin:
Closed in favor of #376.

PR #363, opened 2021-06-22
Refactor Client API

reidrankin:
This is a possible alternative to #361, with a more-thorough refactoring (which only covers `xchain-client` and `xchain-cosmos` at the moment).

The key insight here is that all the clients' exposed API surfaces consist of two types of things:
- Getters and setters which handle configuration and return synchronously; this data is now encapsulated in a per-`Network` `ClientParams` object.
- Asynchronous methods which actually do stuff, and which access but do not mutate the configuration. These have been factored out into a `Delegate` interface, all of whose methods are passed the currently-active `ClientParams` when called and return `Promise`s.

Most of the "computation" done by the existing synchronous getters and setters consists of simply switching behavior based on whether `mainnet` or `testnet` is selected, and this approach eliminates a lot of that complexity. The core "glue" code is contained in the `Client` class, which is generic and can stitch together arbitrary delegates and parameter types. This means that 95% of the code in each dependent package can go in a delegate -- whose only state is the mnemonic phase (or other arbitrary information needed to unlock the client -- like, for example, an instance of `@shapeshiftoss/hdwallet-native`).

I'd love to go over this in more detail... please make liberal use of the per-line comment functionality!

### DIFF
--- a/packages/xchain-binance/__tests__/client.test.ts
+++ b/packages/xchain-binance/__tests__/client.test.ts
@@ -74,71 +74,62 @@ describe('BinanceClient Test', () => {
   const testnetClientURL = 'https://testnet-dex.binance.org'
 
   beforeEach(async () => {
-    bnbClient = new BinanceClient({ phrase, network: 'mainnet' })
+    bnbClient = await BinanceClient.create({}, 'mainnet', phrase)
   })
 
   afterEach(async () => {
-    bnbClient.purgeClient()
+    await bnbClient.purgeClient()
   })
 
   it('should start with empty wallet', async () => {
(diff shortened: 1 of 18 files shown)


reidrankin:
(I'm still on the fence about how the signature of `BoundClientFactory` -- and thus `Client.create()` -- should look. Feedback welcome.)

reidrankin:
For reference, [here's](https://gist.github.com/reidrankin/85d3884e28418a12a82efdb521099233) a gist containing a summary of the current publicly-exposed API, in the simplest-possible representation.

reidrankin:
Closed in favor of #376.

PR #370, opened 2021-06-27
adding fee options to client interface such that one can call it in a…

severinwullschleger:
… generic way:

e.g.
if you get the client in a generic way and want to pass a customized gasprice in case it is ethereum
```
this.getClientFromAsset(asset).transfer({
    asset,
    recipient,
    amount: assetToBase(assetAmount),
    gasPrice,
});
```

### DIFF
--- a/packages/xchain-client/src/types.ts
+++ b/packages/xchain-client/src/types.ts
@@ -1,5 +1,5 @@
 import { Asset, BaseAmount } from '@xchainjs/xchain-util'
-import { BigNumber } from 'ethers'
+import { BigNumber } from 'bignumber.js'
 
 export type Address = string
 
--- a/packages/xchain-client/src/types.ts
+++ b/packages/xchain-client/src/types.ts
@@ -1,4 +1,5 @@
 import { Asset, BaseAmount } from '@xchainjs/xchain-util'
+import { BigNumber } from 'ethers'
 
 export type Address = string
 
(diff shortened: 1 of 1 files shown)


packages/xchain-client/src/types.ts
```diff
@@ -57,6 +58,13 @@ export type TxParams = {
   memo?: string // optional memo to pass
 }
 
+export type FeeOptions = {
+  feeOptionKey?: FeeOptionKey
+  feeRate?: number
+  gasPrice?: BaseAmount
+  gasLimit?: BigNumber
```
veado, line 65: It's not generic, since `gasLimit` depends on `ethers.BignNumber`. Or in other words: `FeeOptions` includes a value, which is needed for `xchain-ethereum` only.

veado (changes requested):

packages/xchain-client/src/types.ts
veado: To workaround it: Avoid import of `ethers` by using `bignumber.js` (we already use it in other `xchain-*` packages)

BERFORE
```ts
import BigNumber from 'bignumber.js'
```
CHANGE
```ts
import { BigNumber } from 'ethers'
```

severinwullschleger: true makes sense.

veado: It seems current implementation of `transfer` in `xchain-ethereum` needs to be updated as well to use a correct `BigNumber`.

Pls check if we can merge `TxParams & FeeOptions` to have one type only: `TxParam`

reidrankin: We can, and I did in #376

severinwullschleger: okey cool, lets put that PR on hold for now then. lets wait for your refactor PR

PR #373, opened 2021-06-30
[xchain-eth] Fix `isApproved` call

veado:
Before `isApproved` returned always `false`, it missed to passed`owner` address to ERC20 contract `allowance` function. This PR fixes it. It will also fixes issues based on failing `allowance` handling in ASGARDEX desktop (see https://github.com/thorchain/asgardex-electron/issues/1595, https://github.com/thorchain/asgardex-electron/issues/1594)

- [x] Parameter object for `call`, `estimateCall`, `isApproved` methods
- [x] Optional `gasLimitFallback` param for `approve` call
- [x] Add `estimateApprovel` to `interface EthereumClient`
- [x] Update tests
- [x] Remove log statements
- [x] Bump `v0.22.0`

### DIFF
--- a/packages/xchain-ethereum/src/client.ts
+++ b/packages/xchain-ethereum/src/client.ts
@@ -547,7 +547,7 @@ export default class Client extends BaseXChainClient implements XChainClient, Et
    */
   isApproved = async ({ contractAddress, spenderAddress, amount }: IsApprovedParams): Promise<boolean> => {
     // since amount is optional, set it to smallest amount by default
-    const txAmount = BigNumber.from(amount?.amount() ?? 1)
+    const txAmount = BigNumber.from(amount?.amount().toFixed() ?? 1)
     const owner = this.getAddress()
     const allowance = await this.call<BigNumberish>({
       contractAddress,
@@ -743,10 +743,7 @@ export default class Client extends BaseXChainClient implements XChainClient, Et
         fast: baseAmount(rates.fast, ETH_DECIMAL),
         fastest: baseAmount(rates.fastest, ETH_DECIMAL),
       }
-    } catch (error) {
-      console.log(error)
-      console.warn(`Error pulling rates from thorchain, will try alternate`)
-    }
+    } catch (error) {}
     //should only get here if thor fails
     return await this.estimateGasPricesFromEtherscan()
   }
--- a/packages/xchain-ethereum/src/client.ts
+++ b/packages/xchain-ethereum/src/client.ts
@@ -35,7 +35,6 @@ import {
   FeeOptionKey,
   FeesParams as XFeesParams,
   BaseXChainClient,
-  FeeRates,
 } from '@xchainjs/xchain-client'
 import { AssetETH, baseAmount, BaseAmount, assetToString, Asset, delay } from '@xchainjs/xchain-util'
 import * as ethplorerAPI from './ethplorer-api'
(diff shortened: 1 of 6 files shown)


packages/xchain-ethereum/src/client.ts
```diff
@@ -584,37 +569,38 @@ export default class Client extends BaseXChainClient implements XChainClient, Et
...
-    walletIndex = 0,
-    spender,
-    sender,
-    feeOptionKey,
+    contractAddress,
+    spenderAddress,
+    feeOptionKey = 'fastest',
     amount,
+    walletIndex = 0,
+    gasLimitFallback,
```
mikewyszinski, line 578: if we change the method params i think we should rather allow the user to specify the gasPrice & gasLimit

if either are NOT supplied  ,they should be calculated using   estimateGasPrices() & estimateApprove()

packages/xchain-ethereum/src/client.ts
```diff
@@ -626,20 +612,26 @@ export default class Client extends BaseXChainClient implements XChainClient, Et
...
+        contractAddress,
+        abi: erc20ABI,
+        funcName: 'approve',
+        funcParams: [spenderAddress, txAmount, { from: this.getAddress() }],
+      })
 
       return gasLimit
     } catch (error) {
+      // return `gasLimitFallback` if available
+      if (gasLimitFallback) return BigNumber.from(gasLimitFallback)
```
mikewyszinski, line 635: if the caller has to provide the gas limitFallback as a param (only to treturn it back),  can we just omit it altogether  and let the caller implement their own fallback logic outside of this call?

example

```typescript
let gasLimit: BigNumber
try {
    gaslimit = ethClient.estimateApprove({args})
}catch (error){
    console.log('oopsie cannot calc gaslimit!')
}
if(!gasLimit){
    gasLimit = fallbackGasLimit
}
```

packages/xchain-ethereum/src/client.ts
veado: It's just on option to set a "fallback" value for `gasLimit` in case `estimateApprove` might fail (we had such case in interfaces in the past - see [discussion at Discord](https://discord.com/channels/838986635756044328/842252939149967382/850203681798357023)), but it's not to override the result of `estimateApprove`. If `estimateGasPrices` will fail, there is already a fallback included to use `getDefaultGasPrices` - no need to set it again.

packages/xchain-ethereum/src/client.ts
```diff
@@ -506,71 +494,68 @@ export default class Client extends BaseXChainClient implements XChainClient, Et
...
-      const txAmount = BigNumber.from(amount.amount().toFixed())
-      const allowance = await this.call<BigNumberish>(0, sender, erc20ABI, 'allowance', [spender, spender])
-      return txAmount.lte(allowance)
-    } catch (error) {
-      return Promise.reject(error)
-    }
+  isApproved = async ({ contractAddress, spenderAddress, amount }: IsApprovedParams): Promise<boolean> => {
+    // since amount is optional, set it to smallest amount by default
+    const txAmount = BigNumber.from(amount?.amount().toFixed() ?? 1)
+    const owner = this.getAddress()
```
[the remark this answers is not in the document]
veado: Will update `isApproved` to accept `walletIndex` ^

packages/xchain-ethereum/src/client.ts
veado: > let the caller implement their own fallback logic outside of this call
Agree. I'm going to remove `limitFallback` param for `estimateApprove`

packages/xchain-ethereum/src/client.ts
mikewyszinski: makes sense

mikewyszinski (approved):
ship it!

PR #374, opened 2021-07-01
Unrequire phrase ethereum

maikelmclauflin:
phrases should not be required for creating the client or reading the blockchain
a conditional was added during the constructor to check if a phrase was passed, and consequently not call `client.setPhrase(` if it is not

~i also noted that the `RootDerivationPaths` was not defined in the eth package~ looks like it was just my editor.


reidrankin:
(#376 should also resolve this issue)

PR #375, opened 2021-07-01
added index for sender

maikelmclauflin:
subtle bug for sender address derivation

### DIFF
--- a/packages/xchain-bitcoincash/src/client.ts
+++ b/packages/xchain-bitcoincash/src/client.ts
@@ -279,7 +279,7 @@ class Client extends UTXOClient {
     const { builder, utxos } = await utils.buildTx({
       ...params,
       feeRate,
-      sender: this.getAddress(),
+      sender: this.getAddress(index),
       haskoinUrl: this.getHaskoinURL(),
       network: this.network,
     })


reidrankin:
(I've picked up this fix in #376)

davidp94 (approved):

veado (commented):
Not sure if I miss something, but it seems this PR just added a `try/catch` block, we don't want anymore. Changes of 57de816fc82e54cb82988a03258466f4af7d3cfb seems to be already in `master`. In this case, let's close this PR.

veado (changes requested):
https://github.com/xchainjs/xchainjs-lib/pull/375#pullrequestreview-700972457

davidp94:
> [#375 (review)](https://github.com/xchainjs/xchainjs-lib/pull/375#pullrequestreview-700972457)

I don't see it there https://github.com/xchainjs/xchainjs-lib/blob/master/packages/xchain-bitcoincash/src/client.ts#L347

davidp94:
I will remove the try catch block

packages/xchain-bitcoincash/src/client.ts
```diff
@@ -358,7 +358,7 @@ class Client extends BaseXChainClient implements BitcoinCashClient, XChainClient
     const tx = builder.build()
     const txHex = tx.toHex()
 
-    return await broadcastTx({
+    return broadcastTx({
```
reidrankin, line 361: They're functionally equivalent, but `return await` gives nicer stack traces.

reidrankin (commented):
I'll merge this this morning after I get #386 in.

davidp94:
Thanks.

Le mer. 7 juil. 2021 à 13:39, Reid Rankin ***@***.***> a
écrit :

packages/xchain-bitcoincash/src/client.ts
reidrankin:
```suggestion
    return await broadcastTx({
```

veado (approved):

reidrankin (approved):
rebased and squashed

PR #383, opened 2021-07-05
Clean up and organize imports

reidrankin:
This is based on #382.

One of the major pain points in merging or rebasing new work is that there's more than one way to add (or remove) an import. This confuses git and can lead to various headaches, from accidental duplication of imports to hairy manual conflict resolutions between versions of files that are functionally identical. Enforcing a canonical format for the import section allows multiple changesets that contain the same edits to be merged automatically.

This PR adds a devDependency on `eslint-plugin-ordered-imports`. This package supports automatic on-save fixup, so there's no significant impact to development workflow.


veado (approved):
Nice clean up

PR #384, opened 2021-07-05
Remove trivial array types

reidrankin:
This is based on #383.

This PR removes trivial array types (i.e. `export type Txs = Tx[]`). These exports add no value, clutter import sections, and don't serve as an effective extension point for a generic container because they refer to specific `Array<T>` types instead of something more generic like an `Iterable<T>`.


veado (approved):
Checked last commit a7c1133, which LGTM!

PR #385, opened 2021-07-06
Various Housekeeping Bits & Baubles

reidrankin:
This is based on #384.

This PR contains various minor changes which improve maintainability, roughly ordered by impact.


packages/xchain-bitcoin/src/types/ledger.ts
```diff
@@ -7,7 +8,7 @@ export type LedgerTxInfo = {
   newTxHex: string
 }
 
-export type LedgerTxInfoParams = Pick<TxParams, 'amount' | 'recipient'> & {
+export type LedgerTxInfoParams = OnlyRequired<TxParams> & {
```
veado, line 11: Could you explain how `OnlyRequired<TxParams>` does the same job like `Pick<TxParams, 'amount' | 'recipient'>` before? _I have checked types `OnlyRequired` + `OnlyRequiredKeys`, but don't understand where we set a constraint to use `amount` and `recipient` only ...

veado (approved):
Checked all commits starting from c3d6ded878a6c1b8f2679ed489b1c39c5870befe, which are looking good to me. _Just a question ..._

packages/xchain-bitcoin/src/types/ledger.ts
reidrankin:
```ts
type OnlyRequiredKeys<T, U = keyof T> = U extends keyof T ? (undefined extends T[U] ? never : U) : never
export type OnlyRequired<T> = Pick<T, OnlyRequiredKeys<T>>
```

The secret magic trick is that conditional types (i.e. `type foo<T> = T extends ... ? ... : ...` are [distributive](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types) over union types.

```ts
// note that, for a required param's key:
OnlyRequiredKeys<TxParams, 'recipient'>
=>  'recipient' extends keyof TxParams ? (undefined extends TxParams['recipient'] ? never : 'recipient') : never
=> undefined extends TxParams['recipient'] ? never : 'recipient'
=> undefined extends Address ? never : 'recipient'
=> undefined extends string ? never : 'recipient'
=> 'recipient'

// and that for an optional param's key:
OnlyRequiredKeys<TxParams, 'memo'>
=>  'memo' extends keyof TxParams ? (undefined extends TxParams['memo'] ? never : 'memo') : never
=> undefined extends TxParams['memo'] ? never : 'memo'
=> undefined extends (string | undefined) ? never : 'memo'
=> never

// so, because conditional types are distributive over union types and `keyof` produces a union type:
OnlyRequired<TxParams>
=> Pick<TxParams, OnlyRequiredKeys<TxParams, keyof TxParams>>
=> Pick<TxParams, OnlyRequiredKeys<TxParams, 'walletIndex' | 'asset' | 'amount' | 'recipient' | 'memo'>>
=> Pick<TxParams, OnlyRequiredKeys<TxParams, 'walletIndex'> | OnlyRequiredKeys<TxParams, 'asset'> | OnlyRequiredKeys<TxParams, 'amount'> | OnlyRequiredKeys<TxParams, 'recipient'> | OnlyRequiredKeys<TxParams, 'memo'>>
=> Pick<TxParams, never | never | 'amount' | 'recipient' | never>
=> Pick<TxParams, 'amount', 'recipient'>
```

(Note that `type OnlyRequiredKeys<T, U = keyof T> = U extends keyof T ? ... : never` is not `OnlyRequiredKeys<T, U extends keyof T = keyof T> = ...`. The difference is that generic types with constraints aren't handled distributively.)

PR #386, opened 2021-07-06
Type and Fee Improvements

reidrankin:
This is based on #385.

- Improvements to `Chain`/`Asset` types
- Convert `TxType`, `Network`, `FeeOption`, and `FeeType` to enums
- Add some fee calculation helpers and factor out redundant fee logic in the UTXO coins


packages/xchain-ethereum/__tests__/client.testnet.test.ts
```diff
@@ -383,15 +383,15 @@ describe('Client Test', () => {
     )
     mock_thornode_inbound_addresses_success(thornodeApiUrl, [
       {
-        chain: 'ETH',
+        chain: 'ETH' as Chain,
```
veado, line 387: Before introducing `enum Chain` we were safe to use 'ETH' or any other "string" defined in`xchain-util/src/chain.const.ts`. Now we have to cast it using `as` (which is unsafe). From my perspective we don't need `enum Chain`.

packages/xchain-binance/__tests__/client.test.ts
```diff
@@ -78,44 +79,44 @@ describe('BinanceClient Test', () => {
   const thornodeMainetClientURL = 'https://thornode.thorchain.info'
 
   beforeEach(async () => {
-    bnbClient = new BinanceClient({ phrase, network: 'mainnet' })
+    bnbClient = new BinanceClient({ phrase, network: 'mainnet' as Network })
```
veado, line 82:
```
'mainnet' as Network
```
^ That's a sign not to use an enum - type casting using `as` is unsafe ...

packages/xchain-binance/__tests__/client.test.ts
```diff
@@ -65,9 +65,9 @@ describe('BinanceClient Test', () => {
   const testnetaddress_path1 = 'tbnb1vjlcrl5d9t8sexzajsr57taqmxf6jpmgaacvmz'
 
   const singleTxFee = baseAmount(37500)
-  const transferFee = { type: 'base', average: singleTxFee, fast: singleTxFee, fastest: singleTxFee }
+  const transferFee = { type: 'base' as FeeType, average: singleTxFee, fast: singleTxFee, fastest: singleTxFee }
```
veado, line 68: Pls don't use `as` type casting ... It's just a sign types don't not work properly.

veado (changes requested):
Started review from 45a48d3e5f79c73ff1171404324b4409751f116c.

No need to use `enum` for strings if we have to type cast strings afterwards using `as MyType` .

For all extracted helpers (which is a nice step) would be great to have tests (by another PR)

packages/xchain-ethereum/__tests__/client.testnet.test.ts
reidrankin: A TypeScript `enum` is just syntactic sugar for an exported type and an exported object with the same name:
```
export enum Foo {
  Bar = 'bar'
  Baz = 'baz'
}
```
is equivalent to
```
export type Foo = 'bar' | 'baz'
export const Foo = {
  Bar: 'bar',
  Baz: 'baz',
}
```

Enums thus provide two distinct advantages over plain union types:

First, they provide not only the union type but a mapping object that exists at runtime Note, for example, how this is used in the `singleFee()` helper:
```
export function singleFee(feeType: FeeType, amount: Fee): Fees {
  return Object.values(FeeOption).reduce<Partial<Fees>>((a, x) => ((a[x] = amount), a), {
    type: feeType,
  }) as Fees
}
```
Here, `Object.values(FeeOption)` won't work if FeeOption is only the name of a type.

Second, the exported object effectively provides a "namespace" for the enum values. For example:
```
enum Foo {
    Foo = 'foo',
    Bar = 'bar',
}

enum Bar {
    Foo = 'foo',
    Bar = 'bar',
}

let foo: Foo = Foo.Foo
foo = Foo.Bar
foo = Bar.Foo // Type 'Bar.Foo' is not assignable to type 'Foo'.ts(2322)
```

This is a good thing, because it makes you be more careful about where you're getting your values from. There are a lot of external libraries using the strings `mainnet` and `testnet` as inputs and outputs, for example, and they may not mean exactly the same thing in each context. Using an enum forces you to either use `Network.Mainnet` or `'mainnet' as Network` -- and either way forces you to be upfront that yes, you know what you're doing, and that you _do_ mean _that_ specific kind of `'mainnet'`.


Note that the casts are only used in `*.test.ts`; everywhere else direct references to the enum members (i.e `Chain.Ethereum` or `Network.Mainnet` or such) have been used; I did this specifically so that the tests would provide assurance that the change does not break the existing API for clients that already use literal strings.

veado: Thanks for this detailed explanation ^. Let's go with `enum` then.

veado (approved):

PR #388, opened 2021-07-07
Prepare bump for `xchain-client|util`

veado:
- [x] `xchain-client@0.10.0`
- [x] `xchain-util@0.3.0`


packages/xchain-client/CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
+# v.0.10.0 (2021-07-07)
+
+### Breaking change
+
+- Introduce `Network`, `TxType`, `FeeOption` enums
+- added support for fetch gas fees from thorchain
```
veado, line 6: Remove that line ^

reidrankin: this is already covered in the 0.9.4 entry
```suggestion
```

reidrankin (approved):
note the comment; otherwise, LGTM

packages/xchain-client/CHANGELOG.md
reidrankin: (lol i was too slow, this done be fixed already)

reidrankin:
oh right you gotta bump the dep versions all around too I guess

veado:
> oh right you gotta bump the `package.json` versions all around too I guess

It will be another (next) PR. `xchain-client|util` need to be published before, in other case all other packages will fail (as we currently have with `xchain-binance`)

veado:
For now we ignore failing tests for `xchain-binance` (see https://github.com/xchainjs/xchainjs-lib/pull/388/checks#step:6:27) - it needs to be updated with latest `xchain-client|util`
```
Error: /home/runner/work/xchainjs-lib/xchainjs-lib/packages/xchain-binance/src/client.ts(18,3): semantic error TS2305: Module '"@xchainjs/xchain-client"' has no exported member 'singleFee'.
```

reidrankin:
that's not it, it's that the deps don't match as specified any more

veado:
> that's not it, it's that the deps don't match as specified any more

Yep, but you can't add dependencies of `xchain-client@0.10.0` / `xchain-util@0.3.0` to `xchain-binance` as long as these are not published or do I miss something?

reidrankin:
I'm pretty sure you can -- I think it's failing because lerna is putting the `xchain-binance` build ahead of the `xchain-client` one, because it's noticed that `"@xchainjs/xchain-client": "^0.9.3"` dep in `xchain-binance` doesn't match the version in `/packages/xchain-client/package.json` so it hasn't realized they have to be built in order.

PR #390, opened 2021-07-07
Prepare `xchain-client@0.10.1`

veado:


veado:
We can ignore failing tests - it will be fixed with next PR (similar to https://github.com/xchainjs/xchainjs-lib/pull/388#issuecomment-875824039)

PR #391, opened 2021-07-08
fix bitcoin cash transfer - only sign utxos which are chosen for input

severinwullschleger:
when testing on mainnet it scanned for UTXOs where the value was 0. So it didn't "choose" it for the input. but the whole UTXOs array was returned to the transfer function. within the transfer function the it tried to sign also the value 0 utxo which resultet in an error.

the change only signs the input UTXOs


packages/xchain-bitcoincash/src/utils.ts
```diff
@@ -298,7 +298,7 @@ export const buildTx = async ({
 
   return {
     builder: transactionBuilder,
-    utxos,
+    inputUTXOs: inputs,
```
severinwullschleger, line 301: line 269 only returns utxos with a value > fee in the inputs array

packages/xchain-bitcoincash/src/client.ts
```diff
@@ -286,7 +286,7 @@ class Client extends UTXOClient {
 
     const keyPair = this.getBCHKeys(this.phrase, derivationPath)
 
-    utxos.forEach((utxo, index) => {
```
severinwullschleger, line 289: here it tried to sign every utxo found at the address, although some are possibly not chosen for the inputs array

veado:
If `scanUTXOs` was broken for `xchain-bitcoincash`, are similar helpers for `xchain-bitcoin|litecoin` are still fine? BTW: I've just added you as a contributor, so you can push your PRs directly into `xchainjs` repo to trigger CI to run tests (it won't work with forks)

PR #403, opened 2021-07-19
fixes for 398, 401, 402

mikewyszinski:
# binance & thorchain

- Updatedrollupjs to include axios to enlable usage on node

# xchain-client

- optimized BaseXchainClient to skip creating an addres that just gets thown away in the constructor

# xchain-cosmos 

- upgraded to     "cosmos-client": "0.42.7"


davidp94 (approved):
lgtm

PR #405, opened 2021-07-19
FIX -rollback  thorchain & cosmos to  "cosmos-client": "0.39.2"

mikewyszinski:
- also including the fix submitted in  https://github.com/xchainjs/xchainjs-lib/pull/407

PR #407, opened 2021-07-26
fix: add missing walletIndex when get signer address

### LINKED ISSUE #406: BNB: unable to sign tx from any wallet where walletIndex gt 0
When trying to send tx from address that was generated with offset greater then 0, I get an error 
```
PubKey of account does not match PubKey of signature
```

That happens because following line is missing `walletIndex`
https://github.com/xchainjs/xchainjs-lib/blob/master/packages/xchain-binance/src/client.ts#L352

That should be like that

```
const transferResult = await this.bncClient.transfer(
   this.getAddress(walletIndex),
   recipient,
   baseToAsset(amount).amount().toString(),
   asset ? asset.symbol : AssetBNB.symbol,
   memo,
)
```

dimitrykislichenko:
Fix #406

### DIFF
--- a/packages/xchain-binance/src/client.ts
+++ b/packages/xchain-binance/src/client.ts
@@ -349,7 +349,7 @@ class Client extends BaseXChainClient implements BinanceClient, XChainClient {
     await this.bncClient.setPrivateKey(this.getPrivateKey(walletIndex || 0))
 
     const transferResult = await this.bncClient.transfer(
-      this.getAddress(),
+      this.getAddress(walletIndex),
       recipient,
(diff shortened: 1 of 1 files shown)


davidp94 (approved):

mikewyszinski (approved):

mikewyszinski:
hanks for finding and submitting a fix. since this is a one liner, i'm going to add it to the https://github.com/xchainjs/xchainjs-lib/pull/405 , which also needs approval (since last 7 days) can @veado and @davidp94 please have a look?

PR #456, opened 2021-12-27
Issue 455/add stagenet support

polaris-9r:
### Description
This PR introduces a new stagenet value for the `Network` enum and stagenet support in the`BaseXChainClient` class in `xchain-client`, to be used in the `xchain-thorchain` for stagenet support. Originally these changes were only limited to the `xchain-thorchain` package, but because this is a breaking change for `xchain-client`, all other packages dependent on the `Network` enum have been updated to handle a stagenet case by resolving to the values for their respective mainnet configuration.

### Notes
With respect to `xchain-binance`, there is a type dependency from the `@binance-chain` JS SDK in the `chooseNetwork` function that enforces only `mainnet` and `testnet` values that requires a workaround until the SDK supports a stagenet value.

### To-dos
- Update CHANGELOG dates when this is ready to merge to today's date


polaris-9r:
Looks like xchain-binance has a `chooseNetwork` method that relies on an enum that doesn't have a 'stagenet'

packages/xchain-binance/src/client.ts
```diff
@@ -59,6 +58,18 @@ export type MultiSendParams = {
...
+ * Binance-specific Network enum that excludes 'stagenet' because
+ * @binance-chain has type checks limited to 'mainnet' and 'testnet',
+ * but `xchain-client` has 'mainnet', 'stagenet', and 'testnet'
+ *
+ * Remove this once @binance-chain has stagenet support.
+ */
+export enum Network {
+  Mainnet = 'mainnet',
+  Testnet = 'testnet',
+}
```
veado, line 71: No need to introduce another `Network` type, just map `stagenet` to `mainnet` internally (as you have already done in xchain-binance's client)

packages/xchain-binance/src/client.ts
```diff
@@ -105,6 +116,25 @@ class Client extends BaseXChainClient implements BinanceClient, XChainClient {
...
+   * Remove this once @binance-chain has stagenet support.
+   * @returns {Network}
+   */
+  getNetwork(): Network {
+    switch (super.getNetwork()) {
+      case Network.Mainnet:
+        return Network.Mainnet
+      case Network.Testnet:
+        return Network.Testnet
+      default:
```
veado, line 132: [nitpick] Handle `stagenet`, but not `default` here and at other `switch/case` statement in this file. `default` might be unsafe in case of other updates in `Network` type.

packages/xchain-bitcoin/src/blockstream-api.ts
```diff
@@ -18,6 +18,9 @@ export const broadcastTx = async ({ network, txHex, blockstreamUrl }: BroadcastT
         return `${blockstreamUrl}/api/tx`
       case Network.Testnet:
         return `${blockstreamUrl}/testnet/api/tx`
+      case Network.Stagenet:
+        // stagenet is not configured, default to mainnet value
+        return `${blockstreamUrl}/api/tx`
```
veado, line 23: Avoid duplication of code by grouping `Network.Mainnet`|`Network.Stagenet` in `switch/case` statements here and at other places. Especially in places with more logic included (e.g. in xchain-ethereum client etc.).

BEFORE:
```ts
switch (network) {
      case Network.Mainnet:
        return ruleA
      case Network.Testnet:
        return ruleB
      case Network.Stagenet:
          return ruleA
```

CHANGE:

```ts
switch (network) {
      case Network.Mainnet:
      case Network.Stagenet:
        return ruleA
      case Network.Testnet:
        return ruleB
```

packages/xchain-client/src/BaseXChainClient.ts
```diff
@@ -18,6 +18,7 @@ import {
 } from './types'
 
 const MAINNET_THORNODE_API_BASE = 'https://thornode.thorchain.info/thorchain'
+const STAGENET_THORNODE_API_BASE = 'https://stagenet-thornode.ninerealms.com'
 const TESTNET_THORNODE_API_BASE = 'https://testnet.thornode.thorchain.info/thorchain'
```
veado, line 22: Ohter endpoints might be changed to `ninerealmscom` as well?

veado (changes requested):
Nice first PR :) Some suggestions ...

packages/xchain-thorchain/__tests__/client.test.ts
```diff
@@ -161,6 +161,10 @@ describe('Client Test', () => {
         node: 'new mainnet client',
         rpc: 'new mainnet client',
       },
+      stagenet: {
```
veado, line 164: Please add a test for `validateAddress` using `stagenet` ([line 152](https://github.com/xchainjs/xchainjs-lib/pull/456/files#diff-e33969a1dd498f30c36ef473117bb07d3244b1783aac289ab366ec292e4255bcL152))

packages/xchain-thorchain/src/util.ts
```diff
@@ -98,6 +98,8 @@ export const getPrefix = (network: Network) => {
   switch (network) {
     case Network.Mainnet:
       return 'thor'
+    case Network.Stagenet:
+      return 'sthor'
     case Network.Testnet:
       return 'tthor'
```
veado, line 104: Pls add a test for `getPrefix` (we don't have it yet)

packages/xchain-client/src/BaseXChainClient.ts
polaris-9r: I believe we have a mainnet api endpoint that we can upgrade to, but not a testnet. Double-checking.

packages/xchain-thorchain/src/util.ts
polaris-9r: Test added in `util.test.ts`. Resolved.

packages/xchain-thorchain/__tests__/client.test.ts
polaris-9r: Test added. Resolved.

packages/xchain-client/src/BaseXChainClient.ts
polaris-9r: API bases updated for mainnet

polaris-9r:
Requested changes have been made @veado

veado (approved):

PR #470, opened 2022-01-25
WIP: Fix bugs

JonnyBurger:


JonnyBurger:
Sorry! Was meant to be against our fork

veado:
@JonnyBurger In case you have found some bugs or issues to solve, could you open a PR to push these to upstream (original xchainjs repo) as well, pls?

JonnyBurger:
@veado 
I this case we wanted to make it work in React Native and the branches have diverged quite a bit. So the bugs I was mentioning were ones we wrote ourselves. But if we find bugs in upstream, we'll send the fixes back!

PR #528, opened 2022-03-31
feat: add tx fee bounds [WIP]

fedorovian:
WIP. Applies fee bounds to BCH, BTC, LTC, DOGE, ETH.

ATOM has hardcoded fees.

BNB, DOT, TERRA and THOR are out of the standard on the fee creation. Will add logic for each of them on the specific place that calculates fees.


thorchain-admin:
Instead of asking the developer to add fee_bounds and passing into `Utils.getDefaultFees()` , can they simply be queried internally. Thus xchainjs community needs to make no changes to their interfaces to inherit this protection.

fedorovian:
https://github.com/xchainjs/xchainjs-lib/pull/554

PR #550, opened 2022-04-11
Unrequire phrase

l0blolly:
Phrases in the ethereum client should not be required to create a client. 
client error

The eth client is the only client where phrase is not optional. `purgeClient()` should remove the phrase. Methods that require a phrase should throw errors if it is not present. 

I've added tests to the other clients to ensure they work the same way. 

I believe this supersedes of #374 

Close #374


veado (approved):
LGTM! Nice to see new tests are included with this PR. Pls bump `xchain-ethereum` (minor version) + update its CHANGELOG

veado (approved):

PR #561, opened 2022-04-28
xchain-util/haven

MartyHav:
This PR adds Haven to xchain-utils ( to enum Chain ). Tests are passing. 
Needed for further haven integration as it is imported into the local client package for haven.


MartyHav:
Doing one PR for complete integration

PR #564, opened 2022-05-05
Fix/556 gas calc

### LINKED ISSUE #556: [xchain-thorchain] Another try to use `gas: auto` + `gasAdjustment: 1.5` (or higher)
to fix `out of gas`. Discusssion https://discord.com/channels/838986635756044328/967336653075132448/967365186774257664

mikewyszinski:
Close #556


packages/xchain-thorchain/src/util.ts
```diff
@@ -235,7 +236,31 @@ export const buildUnsignedTx = ({
...
   return new cosmosclient.TxBuilder(cosmosSdk, txBody, authInfo)
 }
+export const getGasExpectedForTx = async ({
+  cosmosSDKClient,
+  txBody,
+  privKey,
+}: {
+  cosmosSDKClient: CosmosSDKClient
+  txBody: proto.cosmos.tx.v1beta1.TxBody
+  privKey: proto.cosmos.crypto.secp256k1.PrivKey
```
veado, line 246: Remove `privKey` and pass `account: proto.cosmos.auth.v1beta1.IBaseAccount` instead. Or `sequence` and `account_number` instead of `account`. This will save request of `cosmosSDKClient.getAccount`, which might been already done by a caller of `getGasExpectedForTx`  (see `client.transfer`)

packages/xchain-thorchain/src/client.ts
```diff
@@ -613,6 +631,32 @@ class Client extends BaseXChainClient implements ThorchainClient, XChainClient {
...
+      if (gas_expected === undefined) {
+        throw new Error('could not estimate gas limit required for TX')
+      }
+      const gasLimit = cosmosclient.Long.fromString(gas_expected).multiply(multiplier).toString()
+
+      return gasLimit
+    } catch (error) {
+      console.error(error)
+      console.warn(`Using the fallbackGasLimit of ${fallbackGasLimit} `)
+      return fallbackGasLimit
```
veado, line 657: Don't return a hard-coded fallback. If `getGasExpectedForTx` fails there is a reason and broadcasting a tx won't work after that. Because `getGasExpectedForTx` includes a call of `buildUnsignedTx` which should not fail.

packages/xchain-thorchain/src/client.ts
```diff
@@ -613,6 +631,32 @@ class Client extends BaseXChainClient implements ThorchainClient, XChainClient {
       return getDefaultFees()
     }
   }
+
+  private async determineGasLmit({
```
veado, line 635: Pls rename `determineGasLmit` to `estimateGasLimit` or even better: Remove this function (see next comment below)

packages/xchain-thorchain/src/client.ts
```diff
@@ -613,6 +631,32 @@ class Client extends BaseXChainClient implements ThorchainClient, XChainClient {
...
+    privKey: proto.cosmos.crypto.secp256k1.PrivKey
+    multiplier: number
+    fallbackGasLimit: string
+  }): Promise<string> {
+    try {
+      const gas_expected = await getGasExpectedForTx({ cosmosSDKClient: this.getCosmosClient(), txBody, privKey })
+      if (gas_expected === undefined) {
+        throw new Error('could not estimate gas limit required for TX')
+      }
+      const gasLimit = cosmosclient.Long.fromString(gas_expected).multiply(multiplier).toString()
```
veado, line 651: ^ This could be already done in `util.getGasExpectedForTx`. And by removing the fallback (see another comment below), determineGasLmit can be removed from client.

veado (changes requested):
LGTM - just some suggestions ...

Remove `package-lock.json`

packages/xchain-thorchain/src/util.ts
```diff
@@ -235,7 +236,31 @@ export const buildUnsignedTx = ({
...
+}): Promise<string | undefined> => {
+  const accAddress = cosmosclient.AccAddress.fromPublicKey(privKey.pubKey())
+  const account = await cosmosSDKClient.getAccount(accAddress)
+  const txBuilder = buildUnsignedTx({
+    cosmosSdk: cosmosSDKClient.sdk,
+    txBody: txBody,
+    gasLimit: DEFAULT_GAS_LIMIT_VALUE, // NOTE: this should not matter since we are just doign an estimate
```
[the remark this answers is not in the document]
mikewyszinski: yep, good idea. done

packages/xchain-thorchain/src/util.ts
```diff
@@ -235,7 +236,31 @@ export const buildUnsignedTx = ({
 
   return new cosmosclient.TxBuilder(cosmosSdk, txBody, authInfo)
 }
+export const getGasExpectedForTx = async ({
```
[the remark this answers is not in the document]
mikewyszinski: done

packages/xchain-thorchain/src/util.ts
```diff
@@ -20,8 +20,9 @@ import { MsgNativeTx } from './types/messages'
 import types from './types/proto/MsgCompiled'
 
 export const DECIMAL = 8
-export const DEFAULT_GAS_VALUE = '4000000'
-export const DEPOSIT_GAS_VALUE = '500000000'
+export const DEFAULT_GAS_ESTIMATE_MULTIPLIER = 2
```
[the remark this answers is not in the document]
mikewyszinski: done

packages/xchain-thorchain/src/util.ts
mikewyszinski: 1. i recall trying to invoke the /simulate endpoint with an unsigned tx and getting some kind of error.
2. the problem is that transferOffline() also calls determineGasLimit() and it doesn't have an account, so i just knowingly made another possibly redundant call to getAccount() inside getGasExpectedForTx() to keep it simple.

we can discuss this one further if you think it's better to try to get rid of it.

packages/xchain-thorchain/src/client.ts
mikewyszinski: hmm interesting...i was thinking if there is some kind of issue with calling /simulate, the tx should at least give it a try   it another try with the back up gas limit? I'll let you respond before removing the  fallback entirely

packages/xchain-thorchain/src/client.ts
mikewyszinski: agreed, if I remove fallback gas limits, this entire method can be removed...so i'll wait until you reply above ;-)

packages/xchain-thorchain/src/client.ts
mikewyszinski: done

veado (dismissed):

Thorian1te (approved):
lgtm

PR #566, opened 2022-05-06
Add getting internal tx for ETH

osabl:
Just getting internal tx for ETH for `getTransaction` method


mikewyszinski:
@osabl  2 general comments:

1. tests are failing on this branch, will need to fix before merging 
2. i'm thinking this change may confuse anyone using this library already because now internal txs will be displayed when they weren't prior. Should we consider NOT combining internalTXs in the `getTransactions()` call and expose `getETHInternalTransactionHistory()` publically instead?

Issue #541, opened 2022-04-04
FIX: -ethereum Hide balances below 4 sigfig


thorchain-admin:
Add a check in `getBalance()` to hide dust balances 

eg <0.0001 is hidden

mikewyszinski:
this was quick to do, but feels like the caller could easily filter any balances they do not want to see themselves, no?

veado:
> this was quick to do, but feels like the caller could easily filter any balances they do not want to see themselves, no?
> 
> not sure this change is needed..

Agree. Let wallet devs add this feature in its wallet code if needed - but not in xchain-ethereum.

mikewyszinski:
ok going to close this without merging, if any caller feels the need to implement this it can be done as follows:

```
export const MIN_TOKEN_AMOUNT_TO_LIST = baseAmount(0.0001, ETH_DECIMAL)
balances.filter((balance) => balance.amount.gt(MIN_TOKEN_AMOUNT_TO_LIST))
```

PR #576, opened 2022-05-17
Fix/541 hide dust balances

mikewyszinski:


veado:
Closed - see https://github.com/xchainjs/xchainjs-lib/issues/541#issuecomment-1128554100 + https://github.com/xchainjs/xchainjs-lib/issues/541#issuecomment-1129010710

PR #588, opened 2022-05-24
add xchain-haven

MartyHav:
This PR adds xchain-haven package. 

a few marks: 

-  xchain-util is included as a local package as the npm version, doesn't include enum Chain.Haven yet
https://github.com/haven-protocol-org/xchainjs-lib/blob/261e0371771a0331fd1489d62f872b0d0f90e90a/packages/xchain-haven/package.json#L37

- mnemonicConverter is included as a GitHub package and set to latest commit. It exposes two functions, where you can either convert a bip39 phrase into a haven or monero mnemonic. It was planed to move this package/functionality to xchain-crypto package and I would like to leave this to xchainjs maintainers. 
https://github.com/haven-protocol-org/xchainjs-lib/blob/261e0371771a0331fd1489d62f872b0d0f90e90a/packages/xchain-haven/package.json#L40

- haven-core-js the underlying wallet library/sdk is included as a GitHub package as well for now. I'll create a npm package of it shortly.

Besides of these points, this PR is ready to be reviewed.

PR #593, opened 2022-05-28
add bip39 converter for haven and monero to crypto package

MartyHav:


packages/xchain-crypto/package.json
```diff
@@ -54,6 +57,7 @@
   "dependencies": {
     "bech32": "^2.0.0",
     "bip39": "^3.0.4",
+    "bip39-converter": "^1.0.3",
```
veado, line 60: @MartyHav Where is the source code of `bip39-converter` published? Could not find any information at https://www.npmjs.com/package/bip39-converter nor at https://github.com/Coinomi/bip39-coinomi/

veado: Just found it https://github.com/haven-protocol-org/bip39mnemonicConverter

Pls add this repo to https://www.npmjs.com/package/bip39-converter

veado (changes requested):
Thanks for your PR! After re-thinking my [previous comment](https://github.com/xchainjs/xchainjs-lib/pull/588#issuecomment-1138752823) I would suggest the following: Since `convertToHavenMnemonic` is just wrapper of `convertBip39ToHavenMnemonic` and used by `xchain-haven` only and not any other `xchain-*` package, it should be moved into `utils` of `xchain-haven`. Monero can do the same by placing it into its `util` of `xchain-monero`.

packages/xchain-crypto/package.json
MartyHav: will be added with next version

MartyHav:
Alright, closing this PR.

PR #597, opened 2022-06-16
Changes to Cosmos client for ATOM

towanTG:
The only problem we encountered with xchainjs, is having api.cosmos.network being CORS restricted.
We therefore implemented our own API into the client.

We also set the default amount for the DEFAULT_FEE  for transactions. Should I include those changes as well?

### DIFF
--- a/packages/xchain-cosmos/package.json
+++ b/packages/xchain-cosmos/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@xchainjs/xchain-cosmos",
-  "version": "0.17.0",
+  "version": "0.17.1",
   "description": "Custom Cosmos client and utilities used by XChainJS clients",
   "keywords": [
     "XChain",
--- a/packages/xchain-cosmos/src/client.ts
+++ b/packages/xchain-cosmos/src/client.ts
@@ -29,14 +29,20 @@ export interface CosmosClient {
   getSDKClient(): CosmosSDKClient
 }
 
-const MAINNET_SDK = new CosmosSDKClient({
-  server: 'https://api.cosmos.network',
-  chainId: 'cosmoshub-4',
-})
-const TESTNET_SDK = new CosmosSDKClient({
-  server: 'https://rest.sentry-02.theta-testnet.polypore.xyz',
-  chainId: 'theta-testnet-001',
-})
+export interface CosmosClientParams extends XChainClientParams {
+  serverUrl?: string
+}
+
+const MAINNET_SDK = (serverUrl: string | undefined) =>
+  new CosmosSDKClient({
+    server: serverUrl || 'https://api.cosmos.network',
+    chainId: 'cosmoshub-4',
+  })
+const TESTNET_SDK = () =>
+  new CosmosSDKClient({
+    server: 'https://rest.sentry-02.theta-testnet.polypore.xyz',
+    chainId: 'theta-testnet-001',
+  })
 
 /**
  * Custom Cosmos client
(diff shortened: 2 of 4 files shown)


towanTG:
Feedback, in any way, very welcome!

packages/xchain-cosmos/src/client.ts
```diff
@@ -29,14 +29,20 @@ export interface CosmosClient {
...
+export interface CosmosClientParams extends XChainClientParams {
+  serverUrl?: string
+}
+
+const MAINNET_SDK = (serverUrl: string | undefined) =>
+  new CosmosSDKClient({
+    server: serverUrl || 'https://api.cosmos.network',
+    chainId: 'cosmoshub-4',
+  })
+const TESTNET_SDK = () =>
```
veado, line 41: [nitpick] Rename `MAINNET_SDK` -> `createMainnetClient` / `TESTNET_SDK` -> `createTestnetClient` or just move initializations of `CosmosSDKClient` into `constructor` of `Client` (it's not needed anywhere else)

packages/xchain-cosmos/src/client.ts
```diff
@@ -29,14 +29,20 @@ export interface CosmosClient {
...
-const MAINNET_SDK = new CosmosSDKClient({
-  server: 'https://api.cosmos.network',
-  chainId: 'cosmoshub-4',
-})
-const TESTNET_SDK = new CosmosSDKClient({
-  server: 'https://rest.sentry-02.theta-testnet.polypore.xyz',
-  chainId: 'theta-testnet-001',
-})
+export interface CosmosClientParams extends XChainClientParams {
+  serverUrl?: string
```
veado, line 33: Add a parameter to set `serverUrl` for all possible `Network`s (including `testnet` etc.). Similar to `xchain-thorchain` or other packages.

packages/xchain-cosmos/src/client.ts
```diff
@@ -119,7 +126,7 @@ class Client extends BaseXChainClient implements CosmosClient, XChainClient {
   }
 
   getSDKClient(): CosmosSDKClient {
-    return this.sdkClients.get(this.network) || TESTNET_SDK
+    return this.sdkClients.get(this.network) || TESTNET_SDK()
```
veado, line 129: Alternative value of `|| TESTNET_SDK()` is an artifact from a previous version. It can be removed it by changing `private sdkClients: Map<Network, CosmosSDKClient>` to `private sdkClients: Record<Network, CosmosSDKClient>`.

packages/xchain-cosmos/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@xchainjs/xchain-cosmos",
-  "version": "0.17.0",
+  "version": "0.17.1",
```
veado, line 3: Let's bump to an alpha version `0.17.1-alpha1` (or by introducing breaking changes `0.18.0-alpha1`) first, so other contributors of projects can add more changes needed for ATOM until we publish an official release. I know someone who is currently on ASGDX on same things as well lol

veado (changes requested):

packages/xchain-cosmos/src/client.ts
veado: Fixed in #599

packages/xchain-cosmos/src/client.ts
veado: Fixed in #599

packages/xchain-cosmos/package.json
veado: Fixed in #599

packages/xchain-cosmos/src/client.ts
veado: Fixed in #599

veado:
Closing in favour of #599

PR #606, opened 2022-07-04
Created new package xchain-thornode, auto generated api - draft

Thorian1te:
Created new package `xchain-thornode`
Auto generated api using openapi
Added relevant exports


packages/xchain-thornode/src/config.ts
```diff
@@ -0,0 +1,4 @@
+export const THORNODE_API_TS_URL = 'https://thornode.thorswap.net/';
+export const THORNODE_API_9R_URL = 'https://thornode.ninerealms.com/';
```
veado, line 2: Let users define custom urls by using `.env`. If it is not set in `.env`, then use `https://thornode.ninerealms.com` by default. All others can be removed from `config.ts`,  `https://thornode.thorchain.info` might be deprecated in the near future.

veado (dismissed):
LGTM - just one comment.

packages/xchain-thornode/src/config.ts
Thorian1te: yep sure thing. ill fix this now

thorchain-admin (commented):
LGTM

PR #610, opened 2022-07-07
convert yaml to json prior to generating the API

mikewyszinski:


packages/xchain-thornode/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@xchainjs/xchain-thornode",
-  "version": "0.1.0-alpha",
+  "version": "0.1.0-alpha2",
```
Thorian1te, line 3: change logs has a different version number  0.0.1.1

Thorian1te (approved):

PR #622, opened 2022-08-08
 Removed package polkadot and chain references in utils & thorchain-amm

Thorian1te:
Updated change-logs


Thorian1te:
Merged into add evm branch.

PR #645, opened 2022-09-28
Revert changes of `assetFromString` + bump packages

veado:
This PR reverts previous changes of `assetFromString` https://github.com/xchainjs/xchainjs-lib/commit/028460a50c2e84f5292332bf528ae6d2899483f1 

Avoid exceptions for any helper functions as they are more expensive to handle, have less performance and lost its type safety. Besides that, by introducing an exception in `assetFromString` other places in code still not handle it properly. To avoid such breaking change, let's introduce another `assetFromStringEx` (`Ex` -> `Exception`) helper, just for the usage of exception and if needed only.

- [x] Revert changes of `assetFromString` 
- [x] Add `assetFromStringEx` - similar to `assetFromString` , but throwing an exception
- [x] Bump all packages to use latest `xchain-util` / `xchain-client`


packages/xchain-midgard/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@xchainjs/xchain-midgard",
```
Thorian1te, line 2: midgard doesn't need a bump

packages/xchain-thornode/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@xchainjs/xchain-thornode",
-  "version": "0.1.0-alpha4",
+  "version": "0.1.0-alpha5",
   "license": "MIT",
```
Thorian1te, line 4: why does Thornode need a bump?

mikewyszinski (approved):

PR #672, opened 2022-10-28
made auth optional

mikewyszinski:


packages/xchain-litecoin/src/node-api.ts
```diff
@@ -12,20 +12,18 @@ import { TxBroadcastResponse } from './types/node-api-types'
...
-  ).data
+  const postData = {
+    jsonrpc: '2.0',
+    method: 'sendrawtransaction',
+    params: [txHex],
+    id: uniqueId,
+  }
+  let response: TxBroadcastResponse
+  if (auth) {
+    response = (await axios.post(nodeUrl, postData, { auth })).data
```
mrnerdhair, line 23: If auth is falsy, won't axios just omit it anyway? If so, you can get rid of the if statement; if not, though, this is fine, so I'll approve preemptively.

mrnerdhair (dismissed):

Thorian1te (approved):
approved

PR #689, opened 2022-11-22
custom headers wip

Thorian1te:


Thorian1te:
Please review and check version numbers.

veado (approved):
Code LGMT, thx!

Question: What about `xchain-midgard|thonode`? Don't we need an update to support custom headers there as well?

Thorian1te:
> Code LGMT, thx!
> 
> Question: What about `xchain-midgard|thornode`? Don't we need an update to support custom headers there as well?

I am not sure, do we? I only updated `thorchain-query/thornode.ts & midgard.ts` for header support.

veado:
> > Code LGMT, thx!
> > Question: What about `xchain-midgard|thornode`? Don't we need an update to support custom headers there as well?
> 
> I am not sure, do we? I only updated `thorchain-query/thornode.ts & midgard.ts` for header support.

Yes, please update these packages as well. As long as an user uses `xchain-midgard` or `xchain-thornode` packages to communicate with 9R servers, they need to send a custom header or it might break.

packages/xchain-thorchain-query/src/utils/midgard.ts
```diff
@@ -45,6 +48,7 @@ export class Midgard {
     this.network = network
     this.config = config ?? defaultMidgardConfig[this.network]
     axiosRetry(axios, { retries: this.config.apiRetries, retryDelay: axiosRetry.exponentialDelay })
+    axios.defaults.headers.common['x-client-id'] = defaultMidgardConfig[this.network].customRequestHeaders
```
veado, line 51: Instead of using global `axios` defaults, which will be applied to EVERY requests, use custom instance defaults. See https://axios-http.com/docs/config_defaults

packages/xchain-thorchain-query/src/utils/thornode.ts
```diff
@@ -63,6 +67,7 @@ export class Thornode {
     this.network = network
     this.config = config ?? defaultThornodeConfig[this.network]
     axiosRetry(axios, { retries: this.config.apiRetries, retryDelay: axiosRetry.exponentialDelay })
+    axios.defaults.headers.common['x-client-id'] = defaultThornodeConfig[this.network].customRequestHeaders
```
veado, line 70: Avoid `using global axios defaults` - see previous comment.

veado (dismissed):

packages/xchain-client/src/BaseXChainClient.ts
```diff
@@ -42,6 +43,7 @@ export abstract class BaseXChainClient implements XChainClient {
     this.chain = chain
     this.network = params.network || Network.Testnet
     this.feeBounds = params.feeBounds || { lower: 1, upper: Infinity }
+    this.customRequestHeaders = params.customRequestHeaders ?? { 'x-client-id': 'xchainjs-client' }
```
veado, line 46: Don't set a default value - see https://github.com/xchainjs/xchainjs-lib/pull/689#discussion_r1032818390

packages/xchain-thorchain-query/src/utils/midgard.ts
Thorian1te: oh okay good catch will fix this shortly

packages/xchain-client/src/BaseXChainClient.ts
mikewyszinski: If we dont set the default value, then if a user of an xchain-* lib does not send in a header value, all requests will fail no?  don't we want a "default"  value so that someone can pull our libs and immediately make requests?

packages/xchain-thorchain-query/src/utils/thornode.ts
mikewyszinski: removed

packages/xchain-thorchain-query/src/utils/midgard.ts
mikewyszinski: removed

packages/xchain-client/src/BaseXChainClient.ts
mikewyszinski: if a user of the lib wants to  turn off the default headers they would need to actively set customRequestHeaders = {}. i think this makes the most sense for onboarding, thoughts?

veado: > If we dont set the default value, then if a user of an xchain-* lib does not send in customRequestHeaders, all requests will fail no? don't we want a "default" value so that someone can pull our libs and immediately make requests?

From #https://github.com/xchainjs/xchainjs-lib/pull/689#discussion_r1032818390:
"Other servers than ninerealms.com might not accept ``'x-client-id'` in headers and can return an CORS error instead."

Thorian1te: > Yes. Just don't set a header in this case to other api endpoints than 9R (see prev. comment [about CORS]

So in this case we do something like this, `this.customRequestHeaders = params.customRequestHeaders ?? { '': '' }`
And in those Clients that use 9R  api endpoints we set the headers to default to this ` { 'x-client-id': 'xchainjs-client' }` ?

mikewyszinski: @veado  have a look a the current impl, i think it does what needs to be done. however, we realized cosmos was missed (when used for making TC requests) and seems like the lib we use doesn't allow any addtional headers to be sent....i created an issue https://github.com/cosmos-client/cosmos-client-ts/issues/105 but I may need to patch the lib or find some other kind of workaround...

veado: > @veado have a look a the current impl, i think it does what needs to be done. however, we realized cosmos was missed (when used for making TC requests) and seems like the lib we use doesn't allow any addtional headers to be sent....i created an issue [cosmos-client/cosmos-client-ts#105](https://github.com/cosmos-client/cosmos-client-ts/issues/105) but I may need to patch the lib or find some other kind of workaround...

Yes, let's try to patch/fix it in a custom fork of `cosmos-client-ts` and use it as `dependency` (not `peerDependency`) in `xchain-*`. Open a PR with this patch, so we can use original `cosmos-client-ts` later again.

mikewyszinski (approved):

packages/xchain-client/src/BaseXChainClient.ts
mikewyszinski: the patch ended up being infeasible so we took the devs suggestion and used the default config, this is less than ideal, but wanted to get support for this completed and we can revisit if anyone reports an issue

PR #694, opened 2022-12-08
changed deposit to use depositWithExpiry

mikewyszinski:


packages/xchain-thorchain-amm/src/utils/eth-helper.ts
```diff
@@ -61,17 +60,19 @@ export class EthHelper {
       }
       const contractAddress = getContractAddressFromAsset(params.asset)
       const checkSummedContractAddress = ethers.utils.getAddress(contractAddress)
+      const expiry = Date.now() / 1000 + FIFTEEN_MIN_IN_SECS
```
veado, line 63: I would not not use computers time here, but `timestamp` provided by latest `block` (https://docs.ethers.io/ethers.js/v3.0/html/api-providers.html?highlight=timestamp#block-tag). Check how ASGDX has implemented `depositWithExpiry` recently https://github.com/thorchain/asgardex-electron/pull/2473/files

mikewyszinski: good catch changed

veado (approved):

PR #695, opened 2022-12-08
[xchain-thornode] Update to latest THORNode@1.101.0

veado:


packages/xchain-thornode/src/generated/thornodeApi/base.ts
```diff
@@ -46,25 +47,15 @@ export interface RequestArgs {
...
-        if (!this.configuration.baseOptions.headers) this.configuration.baseOptions.headers = {}
-        if (!this.configuration.baseOptions.headers['x-client-id'])
-          this.configuration.baseOptions.headers['x-client-id'] = 'xchainjs-client'
-      }
+    constructor(configuration?: Configuration, protected basePath: string = BASE_PATH, protected axios: AxiosInstance = globalAxios) {
+        if (configuration) {
+            this.configuration = configuration;
+            this.basePath = configuration.basePath || this.basePath;
+        }
     }
```
mikewyszinski: this section is supposed to get re-added during the generation script, i re-ran the "generate:types" script  In any case, i'm going to create a new PR to change the x-client-id header to the interceptor implementation you came up with, sho this will all be rolled back anyway...

mikewyszinski (approved):

PR #698, opened 2022-12-13
696 reimplement custom headers

mikewyszinski:


mikewyszinski:
@veado  @Thorian1te  i need to update the readme's with howto info and remove some logging , but other than that this one is done.... 

if you want to see the 2e2 test i added print out  that you can see by running the following `(cd packages/xchain-thorchain-amm && yarn e2e -t 'wallet Tests')`

packages/xchain-cosmos/src/util.ts
```diff
@@ -1,5 +1,5 @@
-import { cosmosclient, proto } from '@cosmos-client/core'
-import { cosmos } from '@cosmos-client/core/cjs/proto'
+// import { cosmosclient, proto } from '@cosmos-client/core'
```
veado, line 1: [nitpick] Can be removed

packages/xchain-thorchain/__tests__/client.test.ts
```diff
@@ -1,4 +1,4 @@
-import { proto } from '@cosmos-client/core'
+import { proto } from '@cosmos-client/core/cjs/module'
```
veado, line 1: Q: Why not `import cosmosclient from '@cosmos-client/core'` (w/o cjs/module) as we have it in other sources?

veado (dismissed):

veado (approved):

PR #719, opened 2023-02-06
Outbound fee calculation bug

Thorian1te:


packages/xchain-thorchain-query/src/thorchain-query.ts
```diff
@@ -229,7 +229,7 @@ export class ThorchainQuery {
         : await this.thorchainCache.convert(params.input, params.input.asset)
 
     const inboundFeeInInboundGasAsset = calcNetworkFee(input.asset, sourceInboundDetails)
-    let outboundFeeInOutboundGasAsset = calcOutboundFee(params.destinationAsset, destinationInboundDetails)
+    let outboundFeeInOutboundGasAsset = calcOutboundFee(params.destinationAsset, destinationInboundDetails).times(3)
```
CodeFunta, line 232: I think that calcNetworkFee must be here instead of calcOutboundFee because outbound fee was taken from thorchain/inbound_addresses endpoint is looks to high as just single inbound fee

mikewyszinski (approved):

PR #726, opened 2023-02-21
Fix: thorchain tx issue

Thorian1te:


packages/xchain-thorchain/src/client.ts
```diff
@@ -382,20 +389,116 @@ class Client extends BaseXChainClient implements ThorchainClient, XChainClient {
...
+      return {
+        hash: txId,
+        asset: senderAsset,
+        from,
+        to,
+        date: new Date(txResult.timestamp),
+        type,
+      }
+    } catch {
+      return await this.getTransactionDataThornode(txId)
```
mikewyszinski, line 450: if this call fails, there is nothing to catch it. should this be here?

Thorian1te: yeah true, I move it out of there

mikewyszinski (approved):

PR #728, opened 2023-02-21
add bsc, fixed checktx

Thorian1te:
Small update to examples folder.


examples/do-swap/doSend.ts
```diff
@@ -28,9 +28,10 @@ const doSend = async (wallet: Wallet) => {
 const main = async () => {
   const seed = process.argv[2]
   const network = process.argv[3] as Network
+  const apiKey = process.env.APIKEY
```
mikewyszinski, line 31: rename to SOCHAIN_API_KEY so we know which one we need

mikewyszinski (changes requested):

mikewyszinski (approved):

Issue #691, opened 2022-11-30
bug: gettransactions doesn't return correct pages 


mikewyszinski:
from discord: https://discord.com/channels/838986635756044328/842252939149967382/1047116790016266250

Does anyone know how to use getTransactions function in xchain-thorchain package to get transactions older than the last 100 txs. I've tried the following snippet:
```
const thorAddress = 'thor1v8ppstuf6e3x0r4glqc68d5jqcs2tf38cg2q6y'
const txHistParams: TxHistoryParams = {
    address: thorAddress, // Address to get history for
    offset: 90, // Optional Offset
    limit: 200, // Optional Limit of transactions
}
const txHistory = await thorClient.getTransactions(txHistParams)
log(txHistory.total)
log(txHistory.txs)
log(txHistory.txs.length)
```

hippocampusSwapper:
While I was working on this issue https://github.com/xchainjs/xchainjs-lib/issues/727 I find the root of this bug, I have made a PR in order to solve it. https://github.com/xchainjs/xchainjs-lib/pull/730

PR #730, opened 2023-02-23
Fix bug gettransactions doesn't return correct pages

hippocampusSwapper:
Issue: https://github.com/xchainjs/xchainjs-lib/issues/691

I have fixed `getTransactions` for xchain-thorchian client, trying to respect the current approach based on offset + limit. 

We can't optimize the number of requests because we need to request outgoing and ongoing transactions separately so we need to ask for all transactions from zero to offset + limit and filter locally.

For that reason, I have added two constants that limit the request that getTransactions can make: 

`MAX_TX_COUNT_PER_FUNCTION_CALL=500`: Maximum number of transactions that each call can return
`MAX_PAGES_PER_FUNCTION_CALL=15`: Maximum number of pages that the function can request

**Note**: Despite getting outgoing and ongoing transaction separately will continue to be a problem. This function could be more straightforward and valuable if we migrate from the offset/limit approach to the page / page-size approach. Currently, the user only can't access the last 1500 transactions.


Thorian1te (commented):
Nice work on the bug fix and the test cases. 
its 1 commit behind master and needs version # in package.json & changelog updated.

mikewyszinski (approved):

Thorian1te (approved):
looks good

PR #731, opened 2023-03-01
723 bug if synth supply over target tx is refunded

Thorian1te:


packages/xchain-thorchain-query/src/utils/thornode.ts
```diff
@@ -378,9 +393,9 @@ export class Thornode {
     toAsset: string,
     amount: number,
     destination: string,
-    toleranceBps: number,
+    toleranceBps: string,
     affiliateBps: number,
-    affiliate: string,
+    affiliate: number,
     height?: number,
```
mikewyszinski, line 399: pretty sure these have not changed types, it looks like the api.quoteSwap() added a new optional param called fromAddress?: string,  this pososibly throws off the other params

(see attached screenshot)

Screen Shot 2023-03-02 at 10 01 17 PM

mikewyszinski (changes requested):

packages/xchain-thorchain-query/src/thorchain-query.ts
```diff
@@ -386,6 +386,13 @@ export class ThorchainQuery {
       const destPool = await this.thorchainCache.getPoolForAsset(destAsset)
       if (!destPool.isAvailable())
         errors.push(`destinationAsset ${destAsset.ticker} does not have a valid liquidity pool`)
+      // check synth info on thornode pools
+      const pools = await this.thorchainCache.thornode.getPools()
+      const destinationAssetPool = pools.find((pool) => pool.asset === `${destAsset.chain}.${destAsset.symbol}`)
```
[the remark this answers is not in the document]
Thorian1te: getPoolForAsset() returns back type <LiquidityPool> which does not have `synth_mint_paused` since it is a MidgardPool fetch. this is why I had to get it directly from thorchain.

packages/xchain-thorchain-query/src/utils/thornode.ts
Thorian1te: good find, thanks

mikewyszinski (approved):

PR #775, opened 2023-06-13
769 fix fall back for access to non pruned nodes, 

Thorian1te:


GrassRootsCrypto (approved):
LTGM. From address was removed and fallback was added with the required config

mikewyszinski (commented):
left a comment, if you think we should make the suggested change, go for it. if not i'll approve later today

mikewyszinski (approved):

PR #784, opened 2023-06-30
updated default gas value, and updated deps in thorchain-amm

Thorian1te:


packages/xchain-thorchain-amm/package.json
Thorian1te, line 1: I don't need to bump thorchain-amm version number in this pr as "version": "0.5.6", has not been release yet.

GrassRootsCrypto (approved):
LGTM

PR #813, opened 2023-08-14
801 add streaming swap example

hippocampus-web3:
Notes about streaming swap @Thorian1te 

- Streaming swap is not compatible with **toleranceBps** (slippage) with the current implementation
- Function `estimateSwap` is not adapted for streaming swap https://github.com/xchainjs/xchainjs-lib/blob/60af8ce17f6dbf8b25a1945d70b5d68cd2425002/packages/xchain-thorchain-amm/src/thorchain-amm.ts#L56

Do we need to create issues to cover this?


Thorian1te (approved):

Thorian1te (approved):
Nice example format

Issue #817, opened 2023-08-26
New Package: Midgard-query


Thorian1te:
Remove Midgard into its own query package. 
Thorchain-query should only depend on THORNode as its data provider.

PR #820, opened 2023-08-31
817 new package Midgard query

hippocampus-web3:
#### Overview:
This PR aims to reduce the coupling between the `xchain-thorchain-query` package and the `xchain-midgard`. The changes introduce a new package called `xchain-midgard-query`, which now handles functionalities previously managed by `xchain-thorchain-query` and depending on `xchain-midgard`.

#### Changes:
1. Moved all `xchain-midgard` dependent functionality to `xchain-midgard-query`.
2. Updated `xchain-thorchain-query` to minimize its dependency on the new `xchain-midgard-query` package.

#### Residual Dependency:
While the goal is to eliminate the dependency, there's one specific operation where this has proven to be challenging. Specifically, querying the number of native decimals for assets in each liquidity pool, which is information not directly provided by THORNode and is only available through Midgard. 

#### Implications:
This change significantly reduces the dependency of `xchain-thorchain-query` on `xchain-midgard-query`, but does not entirely eliminate it. For now, the residual dependency remains, but it's minimal and should not pose rate-limiting issues on Midgard for users operating `XchainJs` against their own THORNodes.

https://github.com/xchainjs/xchainjs-lib/issues/817


packages/xchain-thorchain-query/package.json
```diff
@@ -43,7 +43,7 @@
   },
   "peerDependencies": {
     "@xchainjs/xchain-client": "^0.14.1",
-    "@xchainjs/xchain-midgard": "^0.5.1",
+    "@xchainjs/xchain-midgard-client": "^0.1.0",
```
Thorian1te, line 46: is the package called `xchain-midgard-client` or `xchain-midgard-query`

Thorian1te (approved):
Great work! tests well

PR #822, opened 2023-09-10
Fix conflict versions thorchain-query / thorchain-amm

hippocampus-web3:


packages/xchain-thorchain-query/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@xchainjs/xchain-thorchain-query",
-  "version": "0.5.0",
+  "version": "0.6.0",
```
Thorian1te, line 3: Any reason why we are bumping version by minor and not by patch ?

hippocampus-web3: Yeah, there is a new parameter on ThorchainCache constructor, that break retro compatibility for that reason i think it's a minor not a patch

Thorian1te: happy days then

Thorian1te (approved):

PR #806, opened 2023-08-04
add xchain-mayachain-query

ecalchot-maya:
Still a work in progress.

PR #826, opened 2023-09-14
bump mayanode for latest changes in openapi spec

ecalchot-maya:
There was an error in a field in the `/mayachain/constants` where the name from the openapi spec didn't match the actual field name returned by the api (`int64_values` instead of `int_64_values`) which messed with the types generated from the openapi-generator. This bump is intented to get that fix and be able to publish #806 that needs this fix to get all tests up and running.


hippocampus-web3:
The PR only bump the version but not fix any bug. Are there any missing commits?

ecalchot-maya:
That is it @hippocampus-web3 , there aren't any bugs/corrections at the moment. The only point is to generate the new types from the openapi spec that corrects the field that I mentioned.

Thorian1te:
To re-iterate what @hippocampus-web3  said we don't push a new version unless there are changes to the code base, and this PR has no changes to the code base.  If there are new types from the openapi spec then generate the new types and push the new spec.

ecalchot-maya:
You guys are right, my bad. The latest commit includes the changes I mentioned. Let me know if I need to make any other adjusment.

Thorian1te:
Please fill the change log in for `xchain-mayanode`. 
 There has been a change to `xchain-midgard` but no changes to the change log or the package version number. 
 either discard changes or add the required info. Run `yarn test` in root folder to check to make sure nothing has broken with the new specs. If you are changing `xchain-midgard` then `xchain-midgard-query` needs checking and updating, and packages that depend on it > `xchain-thorchain-query` & `xchain-thorchain-amm`. 
 cc @hippocampus-web3

ecalchot-maya:
Done @Thorian1te , I removed the changes from the midgard package.

Thorian1te (approved):

Issue #846, opened 2023-10-05
CHANGE: UTXO client to default spendPendingUTXOs = true


thorchain-admin:
# Context

UTXO chains can allow users to spend pending UTXOs (still in the mempool) using child-pays-for-parent rules
https://bitcoinops.org/en/topics/cpfp/

Centralised services don't like pending UTXOs because of re-org, double-spend attacks etc, so typically they ignore. This is why API services allow to filter out pending. 

A self-custodial wallet should not be concerned about the confirmation status of a UTXO, since the user and their peers has custody of their own funds. Thus it should be allowed to spend. 

## THORChain
In past TC was sensitive to pending UTXOs because if the fee rate was low, it could be dropped by sending to an inactive vault -- it only accepts confirmed TXs. 

BUT, TC now has "Refunds to Inactive Vaults" so in the low chance a user spends a pendingUTXO with low gas rate into TC and takes more than 3 days to confirm, it will just get refunded. 

# Change

https://github.com/xchainjs/xchainjs-lib/tree/master/packages/xchain-bitcoin/src#L211C5-L211C77
```
spendPendingUTXO = true, // default: spend unconfirmed UTXOs to improve user experience
```

Issue #845, opened 2023-10-03
BTC Client won't sent max amount


GrassRootsCrypto:
Trying to send max balance (over $200 USD) and get an issue "Insufficient Balance for transaction"

sending 692570 BTC to bc1xxx
Error: Insufficient Balance for transaction
    at Client.<anonymous> (/home/wslcmd/gitRepos/xchainjs/xchainjs-lib/examples/do-swap/node_modules/@xchainjs/xchain-bitcoin/src/client.ts:243:36)
    at Generator.next (<anonymous>)
    at fulfilled (/home/wslcmd/gitRepos/xchainjs/xchainjs-lib/examples/do-swap/node_modules/@xchainjs/xchain-bitcoin/lib/index.js:55:58)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)

Issue #816, opened 2023-08-25
Issue with xchain-litecoin and xchain-dogecoin


sectionshankar:
![image](https://github.com/xchainjs/xchainjs-lib/assets/38269146/de825510-233b-44fa-8cbe-39b698d66269)

I often receive these error in spite of having enough balance, most of the time with litecoin.
And also receiving no provider able to get GetUnspentTx

Thorian1te:
@hippocampus-web3 
are you able to have a look at this one pls

hippocampus-web3:
Yes, let me review it @Thorian1te

hippocampus-web3:
Hello @sectionshankar,

Which data provider are you using for Litecoin and Dogecoin?

The error message `no provider available for GetUnspentTx` suggests that there's an issue with the data provider, probably the rate limits on their endpoints on the free tier.

By default, if you don't specify a provider when initializing the client, you'll be using Blockcypher. If you're not using a paid API key, you might be hitting their rate limit. You can review the [limitations](https://www.blockcypher.com/pricing.html) of their service here.

You have three options:

1. Modify your code to reduce the number of requests per second/day to stay within the free tier's rate limits.

2. Purchase a Blockcypher API key with a higher rate limit. Note that Xchain allows you to set a custom API key for data providers.

3. Alternatively, you can use [Sochain](https://sochain.com/api/) in place of Blockcypher. However, keep in mind that they don't offer a free tier, so you'll need to purchase an API key.

Please let me know if this clarifies things for your situation. If you need assistance with any of these solutions, don't hesitate to reach out.

hippocampus-web3:
I closed the issue, if you have any problem related to that don't hesitate to reopen it. Thanks! @sectionshankar

hippocampus-web3:
I think the reason is the same, the problem with rate limit in API providers but the error message 'Insufficient balance' is not descriptive we will improve the error message for that case

sectionshankar:
@hippocampus-web3 @Thorian1te any work on this? Fixing or improving the message?

hippocampus-web3:
Thanks for the reminder. Let me reopen this issue and address an improvement on the messages right now

hippocampus-web3:
[...]

PR #847, opened 2023-10-05
816 bug insufficient funds

hippocampus-web3:
**Fixes a Bug with the Blockcypher Provider**

Resolves an issue where pending UTXO requests were not processed accurately due to the incorrect handling of the appropriate Blockcypher API response field. (https://github.com/xchainjs/xchainjs-lib/issues/845
https://github.com/xchainjs/xchainjs-lib/issues/816)

**Fixes a Bug with the Haskoin Provider**

Corrects a condition where every UTXO was consistently treated as confirmed because the filtering condition to determine its confirmation status was inaccurate.

**Additionally**, the default behavior has been adjusted to always count UTXOs in a pending status, as THORChain no longer possesses a limitation in this aspect. (https://github.com/xchainjs/xchainjs-lib/issues/846)


Thorian1te (dismissed):
Looks good to me, Amm needs a version bump though

Thorian1te (approved):
Great work

Issue #850, opened 2023-10-06
Add parameter 'estimateWithMaxPrecision'


hippocampus-web3:
**Action**: Support a new parameter, 'estimateWithMaxPrecision' for function `getFees` and `getFeesWithMemo` on UTXOClient.

**Rationale**: This allows for precise fee estimation without overburdening the UTXO data provider with excessive requests, maintaining a balance between accuracy and resource utilization. Interfaces can enable this parameter only when a user opts to send the max amount, thereby not drastically increasing the requests to node providers.

Background: https://github.com/xchainjs/xchainjs-lib/issues/845

> The issue arises because we are passing an empty array of inputs for the estimation: https://github.com/xchainjs/xchainjs-lib/blob/4860de136791b490b36c1df81b63e4ca0de57589/packages/xchain-bitcoin/src/utils.ts#L115. Consequently, the estimation is not accurate. The ideal option is to compute the exact number of inputs for each estimation, but this is data-intensive for providers, hence the introduction of a new parameter, 'estimateWithMaxPrecision'. This allows the interface to determine the load it wants to send to their node provider, depending on their tier.
> 
> So, I propose the following plan of action: Temporarily increase the number of UTXOs that we are using for the estimation and, in a subsequent version, refactor to add the parameter 'estimateWithMaxPrecision'. This will likely affect the uxtoClient interface, so it’s crucial to proceed with caution. I believe it's prudent to proceed in this manner to provide you with a working solution in Asgardex right now, without having to wait for 'estimateWithMaxPrecision' to reach, what do you think ?

PR #862, opened 2023-10-15
Estimate fee for sending MAX amount on UTXOs

hippocampus-web3:
Issue: https://github.com/xchainjs/xchainjs-lib/issues/850

Affected: UTXO clients

What Does the PR Do?

1. Moves the logic of the `calcFee` function from individual clients to the `UTXOClient` class.
2. Creates two new protected methods for UTXOClient: `compileMemo` and `getFeeFromUtxos`.
3. Updates signatures for: `getFeesWithRates` and `getFees`.
4. Removes the following functions from utils in all UTXO clients: `getFee`, `calcFee`, `getDefaultFeesWithRates`, and `getDefaultFees`.
5. Removes the deprecated function `getFeesWithMemo`.

Why?

1. (and 2.) The chain-specific logic related to the calcFee process is isolated in compileMemo and `getFeeFromUtxos`. Both are implemented for each chain, but consumed under the same logic in the `calcFee` method.

3. The functions now support a new option sender. If you pass the sender parameter, the function will return the fee for a transaction that includes all UTXOs from this sender (including pending ones). Calling these methods with the option sender is intended for when you want to estimate the commission for sending the maximum available in the wallet. It shouldn't be used in other scenarios because it estimates a high fee and is costly in terms of node provider requests.
 
4. The functions have been removed from utils because you can get fees and fee rates using `getFeesWithRates` and `getFees` directly using the client, instead of importing utils.
 
5. This function has been deprecated for a long time. You can replicate the same functionality using` getFees({ memo: 'ss' })`.


Thorian1te (approved):
Nice update to Fees. looks good

PR #865, opened 2023-10-19
 standardise response in thorchain-query & update thornode to the lat…

Thorian1te:
…est api spec


hippocampus-web3 (approved):

hippocampus-web3:
Let's go 🔥

Issue #876, opened 2023-10-27
Fix: Dogecoin fee bounds are incompatible with the actual blockchain estimation. 


hippocampus-web3:
Context: https://discord.com/channels/838986635756044328/842252939149967382/1167152240650354688

0xp3gasus:
I can take care of it

PR #878, opened 2023-10-27
DOGE Lower fee bound updated

0xp3gasus:
#876

@hippocampus-web3


packages/xchain-doge/src/client.ts
```diff
@@ -161,8 +161,13 @@ class Client extends UTXOClient {
...
   async transfer(params: TxParams & { feeRate?: FeeRate }): Promise<TxHash> {
-    const feeRate = params.feeRate || (await this.getSuggestedFeeRate())
-    checkFeeBounds(this.feeBounds, feeRate)
+    let feeRate: number
+    if (params.feeRate) {
+      feeRate = params.feeRate
+      checkFeeBounds(this.feeBounds, feeRate)
+    } else {
+      feeRate = await this.getSuggestedFeeRate()
+    }
```
hippocampus-web3, line 170: We shouldn't bypass `checkFeeBounds` functionality. If the `LOWER_FEE_BOUND` is too much, update it with a lower quantity.

If the DOGE network does not recommend less than that fee_rate then force a minimum of LOWER_FEE_BOUND.

hippocampus-web3 (approved):

Issue #877, opened 2023-10-27
FIX: 


Thorian1te:
This was a sanity check to see if the asset was parsed in the with correct baseAmount decimals. Since Thornode started offering quotes it was removed as core devs wanted UI's to do a check before sending the quote. 
In any case there should a check before calling estimate swap or doSwap() in amm. 
``` private async isValidSwap(params: EstimateSwapParams) {
    if (isAssetRuneNative(params.input.asset) || params.input.asset.synth) {
      if (params.input.baseAmount.decimal !== 8)
        throw Error(`input asset ${assetToString(params.input.asset)}  must have decimals of 8`)
    } else {
      const assetPool = await this.thorchainCache.getPoolForAsset(params.input.asset)
      const nativeDecimals = assetPool?.pool.nativeDecimal
      if (
        nativeDecimals &&
        nativeDecimals !== '-1' &&
        params.input.baseAmount.decimal !== Number(assetPool?.pool.nativeDecimal)
      ) {
        throw Error(
          `input asset ${assetToString(params.input.asset)}  must have decimals of ${assetPool?.pool.nativeDecimal}`,
        )
      }
    }```

Thorian1te:
This was fixed in the latest release. closing

PR #897, opened 2023-11-05
877 Decimal sanity check

0xp3gasus:
#877 

@Thorian1te


Thorian1te (approved):
Thank you, that is much cleaner

Issue #902, opened 2023-11-07
FIX: Some addresses are failing when calling prepareTx() for cosmos chains


hippocampus-web3:
Error: `Transfer failed - missing pub key`

Error throw in line: 

https://github.com/xchainjs/xchainjs-lib/blob/6cb6a1311cd0c5ce7daaca8478398f18dcce24e1/packages/xchain-mayachain/src/client.ts#L727

PR #903, opened 2023-11-07
Fix prepareTx for uninitialize addresses

hippocampus-web3:
https://github.com/xchainjs/xchainjs-lib/issues/902


Thorian1te (approved):
Looks good to me

Issue #923, opened 2023-11-17
Improve fee calculation strategy on Bitcoin


hippocampus-web3:
Improve the way feeRate is calculated. Now we use for everything the one returned by THORChain. We should be able to choose if the estimation is for interacting with a protocol or for a normal use as we do in xchain-evm and in case of normal use make a calculation from an external provider or from the RPC of the chain.

PR #924, opened 2023-11-19
923 UTXO fee strategy

0xp3gasus:
Issue related #923 

Abstract:

- **Bitgo**: New UTXO provider. Just get fee rates endpoint developed to be focus on the task
- **Blockcypher**: Get fee rates endpoint developed
- **UTXO Fee strategy**. Fees are now calculated from provider by default. To use Thorchain fees to interact with the protocol, use the getFeeRates with Protocol param
- **BTC** upper bound duplicated
- **LTC** lower bound reduced


Thorian1te (commented):
Looks good to me

packages/xchain-thorchain-amm/src/wallet.ts
```diff
@@ -374,6 +416,23 @@ export class Wallet {
...
+        memo: params.memo,
+        feeRate: feeRates.fast,
+      }
+      try {
+        const hash = await assetClient.transfer(addParams)
+        return { hash, url: assetClient.getExplorerTxUrl(hash) }
+      } catch (err) {
+        const hash = JSON.stringify(err)
+        return { hash, url: assetClient.getExplorerAddressUrl(await assetClient.getAddressAsync()) }
+      }
```
hippocampus-web3, line 435: Is there any way to standardize this for all chains? I think this goes against the right way, the trend should be to eliminate these conditionals. In the case of EVMs they are 'indispensable' because the operation of TC demands it, but in this case the conditional is only motivated by the architecture of Xchain. We could standardize it for all strings as much as possible ? if not in this PR at least leave an issue created to eliminate this conditional. Over time this code is difficult to maintain.

0xp3gasus: I agree with you. Let's find out a way to standardise this code, but I think it is better to do it in a different issue, this pull request is quite big

hippocampus-web3 (approved):

PR #942, opened 2023-12-03
Support ClientLedger for BTC

hippocampus-web3:
It exposes a client ready to manage a BTC wallet via Ledger wallet with the same interface as a Keystore wallet.

The idea is that customers such as Asgardex that are currently using parts of the keystore client and must also implement the specific part of integration with Ledger can directly use the Xchain library without worrying about the specific integrations of each wallet. 

Code that could be replace with this package in Asgardex: https://github.com/asgardex/asgardex-desktop/tree/develop/src/main/api/ledger/bitcoin

At the moment the support is only available for BTC, once integrated in Asgardex we will be able to discover possible improvements that are necessary to increase the developer's experience using Xchain.

If this is well accepted, it is not complicated to **export it not only to other types of chains but also to other types of wallets**. **Both hardware and software**

Example: 

```typescript
import TransportNodeHidSingleton from '@ledgerhq/hw-transport-node-hid-singleton'
const transport = await TransportNodeHidSingleton.open()
const btcClient = new ClientLedger({
  transport,
  ...defaultBTCParams,
})

// With BTC app open in ledger
const address = await btcClient.getAddressAsync()
```


Thorian1te (dismissed):
yea boii looks good to me

PR #1029, opened 2024-02-02
1024 cosmos refactor

0xp3gasus:


packages/xchain-cosmos/__e2e__/cosmos-client.e2e.ts
```diff
@@ -1,119 +1,96 @@
...
+    console.log(
+      balances.map((balance) => {
+        return {
+          asset: assetToString(balance.asset),
+          amount: baseToAsset(balance.amount).amount().toString(),
+        }
+      }),
+    )
   })
-  it('Prepate transaction, sign externally and broadcast', async () => {
```
hippocampus-web3, line 72: Can we keep this test ? Some of the leading integrations use this system for signing because they do not work with private keys and it is their only option. Now that .getSDKClient() is not available, their way of signing will be different.

hippocampus-web3 (approved):

PR #1050, opened 2024-02-19
1049 Add Trade Asset Support

GrassRootsCrypto:
Basic support for Trade Assets - https://gitlab.com/thorchain/thornode/-/merge_requests/3433

- Added `trade` in Asset Type
- Updated Asset declarations 
- Updated test cases
- Updated AssetToString with conversion logic

### DIFF
--- a/packages/xchain-util/src/asset.ts
+++ b/packages/xchain-util/src/asset.ts
@@ -204,7 +204,7 @@ export const isSynthAsset = ({ synth }: Asset): boolean => synth
 export const isTradeAsset = ({ trade }: Asset): boolean => trade
 
 const SYNTH_DELIMITER = '/'
-const TRADE_DELIMITER = '-'
+const TRADE_DELIMITER = '~'
 const NON_SYNTH_DELIMITER = '.'
 
 /**
@@ -228,10 +228,10 @@ export const assetFromString = (s: string): Asset | null => {
   const isSynth = s.includes(SYNTH_DELIMITER)
   let isTrade = s.includes(TRADE_DELIMITER)
 
-  isSynth ? true : isTrade = false; // Can't be a Synth and Trade at the same time. 
+  isSynth ? true : isTrade = false; // Can't be a Synth and Trade at the same time.
 
   // const delimiter = isSynth ? SYNTH_DELIMITER : NON_SYNTH_DELIMITER
-   const delimiter = isSynth ? SYNTH_DELIMITER : (!isSynth && !isTrade ? NON_SYNTH_DELIMITER : TRADE_DELIMITER); // Need to look for non-synth or trade, then synth or trade to correctly evaluate 'BNB.RUNE-B1A'  
+   const delimiter = isSynth ? SYNTH_DELIMITER : (!isSynth && !isTrade ? NON_SYNTH_DELIMITER : TRADE_DELIMITER);
 
   const data = s.split(delimiter)
   if (data.length <= 1 || data[1]?.length < 1) {
--- a/packages/xchain-util/src/asset.ts
+++ b/packages/xchain-util/src/asset.ts
@@ -226,17 +226,11 @@ const NON_SYNTH_DELIMITER = '.'
  */
 export const assetFromString = (s: string): Asset | null => {
   const isSynth = s.includes(SYNTH_DELIMITER)
-  let isTrade = s.includes(TRADE_DELIMITER)
+  const isTrade = !isSynth && s.includes(TRADE_DELIMITER) // Ensure isTrade is only considered if not a synth
 
-  isSynth ? true : (isTrade = false) // Can't be a Synth and Trade at the same time.
-  let delimiter
-  if (isSynth) {
-    delimiter = SYNTH_DELIMITER
-  } else if (isTrade) {
-    delimiter = TRADE_DELIMITER
-  } else {
-    delimiter = NON_SYNTH_DELIMITER
-  }
+  // Determine the delimiter based on whether the asset is a synth, trade, or neither
+  const delimiter = isSynth ? SYNTH_DELIMITER : isTrade ? TRADE_DELIMITER : NON_SYNTH_DELIMITER
+
   const data = s.split(delimiter)
   if (data.length <= 1 || data[1]?.length < 1) {
     return null
--- a/packages/xchain-util/src/asset.ts
+++ b/packages/xchain-util/src/asset.ts
@@ -228,11 +228,15 @@ export const assetFromString = (s: string): Asset | null => {
   const isSynth = s.includes(SYNTH_DELIMITER)
   let isTrade = s.includes(TRADE_DELIMITER)
 
-  isSynth ? true : isTrade = false; // Can't be a Synth and Trade at the same time.
-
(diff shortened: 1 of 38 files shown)


packages/xchain-util/src/asset.ts
```diff
@@ -217,7 +226,13 @@ const NON_SYNTH_DELIMITER = '.'
  */
 export const assetFromString = (s: string): Asset | null => {
   const isSynth = s.includes(SYNTH_DELIMITER)
-  const delimiter = isSynth ? SYNTH_DELIMITER : NON_SYNTH_DELIMITER
+  let isTrade = s.includes(TRADE_DELIMITER)
+
+  isSynth ? true : isTrade = false; // Can't be a Synth and Trade at the same time.
```
Thorian1te, line 231: not proper use of the ternary operator. 
you can directly set isTrade to false inside an if statement when isSynth is true
`if (isSynth) {
  // If it's a Synth, it can't be a Trade. Directly enforce this logic.
  isTrade = false;
}`

packages/xchain-util/src/asset.ts
```diff
@@ -217,7 +226,17 @@ const NON_SYNTH_DELIMITER = '.'
  */
 export const assetFromString = (s: string): Asset | null => {
   const isSynth = s.includes(SYNTH_DELIMITER)
-  const delimiter = isSynth ? SYNTH_DELIMITER : NON_SYNTH_DELIMITER
+  let isTrade = s.includes(TRADE_DELIMITER)
+
+  isSynth ? true : (isTrade = false) // Can't be a Synth and Trade at the same time.
```
Thorian1te, line 231: The logic isSynth ? true : (isTrade = false) is problematic. It seems intended to ensure that an asset cannot be both a synth and a trade, but the implementation is flawed.

this is a refactored version 
export const assetFromString = (s: string): Asset | null => {
  const isSynth = s.includes(SYNTH_DELIMITER);
  const isTrade = !isSynth && s.includes(TRADE_DELIMITER); // Ensure isTrade is only considered if not a synth

  // Determine the delimiter based on whether the asset is a synth, trade, or neither
  const delimiter = isSynth ? SYNTH_DELIMITER : (isTrade ? TRADE_DELIMITER : NON_SYNTH_DELIMITER);

  const data = s.split(delimiter);
  if (data.length <= 1 || data[1]?.length < 1) {
    return null; 
  }
  return { chain, symbol, ticker, synth: isSynth, trade: isTrade };
};

packages/xchain-util/src/types/asset.ts
```diff
@@ -4,4 +4,5 @@ export type Asset = {
   symbol: string
   ticker: string
   synth: boolean
+  trade: boolean
```
Thorian1te, line 7: im not sure if this is the correct way to go about adding a trade type to Asset type. @0xp3gasus, Thoughts? 
consider. 

A trade asset cannot be a synth Asset or vice versa. Currently only TC uses trade Assets.

GrassRootsCrypto: It is similar. see https://gitlab.com/thorchain/thornode/-/merge_requests/3326/diffs#e906cd37ce6a35a80301a5eeef02d48cea003cd7_120_123. 

There is an asset type and it can be either:
L1, Synth, Trade or derived.

packages/xchain-util/src/asset.ts
GrassRootsCrypto: Trunk didn't like it either. 
Updated with an if else statement for clarity.

packages/xchain-util/src/types/asset.ts
Thorian1te: yes but it does not inherently support invalid state combinations 
A union type might work better here as with union types, inherently prevents invalid configurations through the type system, offering better safety and explicitness at the cost of increased complexity. 
example 
```
type BaseAsset = {
  chain: string;
  symbol: string;
  ticker: string;
};

type TradeAsset = BaseAsset & {
  synth: false;
  trade: true;
};

type SynthAsset = BaseAsset & {
  synth: true;
  trade: false;
};

type L1Asset = BaseAsset & {
  synth: false;
  trade: false;
};

type Asset = TradeAsset | SynthAsset | L1Asset;
```

Thorian1te: since it is a big change to the asset type,  I would like @0xp3gasus & @hippocampus-web3 to weigh in on this & options

0xp3gasus: My concerns about the pull request. I think Stormzy approach is the one we have to follow but I would not stop just creating the `Asset` type as the union as the three different assets.
If we create this Asset type as the union, there are points in the library where we are going to work with an asset type that maybe is not support. I don't know much info about the aim that `TradeAssets` have so I don't know if this example is correct, but it can be used to understand what I mean. Following this nomenclature, nowadays, a swap can be done with `L1Assets` and `SynthAssets` using Thornode quote swap endpoint, will this endpoint work with `TradeAssets` too? If the answer is yes, this feature can work with the `Asset` type, but, if the answer is no, this feature can not work with the `Asset` type and has to work only with `L1Assets` and `SynthAssets`. Due to the big impact this change has, it's important to define what can and what can not be done with this new type of asset.

Thorian1te: agree with Oxp3gasus. Lets update thornode first and then see what features we can add to xchainjs with this trade account stuff.

0xp3gasus: Thornode client has been updated to v1.128.1 in #1073

Thorian1te: [TradeAccounts](https://dev.thorchain.org/concepts/trade-accounts.html)

```Trade accounts provide professional traders (mostly arbitrage bots) a method to execute instant trades on THORChain without involving Layer1 transactions on external blockchains. Trade Accounts creates a new type of asset, backed by the network security rather than the liquidity in a pool (Synthetics), or by the RUNE asset (Derived Assets).```

* also trade acc are not activated yet 

Since this asset type is not for the average user or interface we should ask ourselves is it worth providing access too.

GrassRootsCrypto: > My concerns about the pull request. I think Stormzy approach is the one we have to follow but I would not stop just creating the `Asset` type as the union as the three different assets. If we create this Asset type as the union, there are points in the library where we are going to work with an asset type that maybe is not support. I don't know much info about the aim that `TradeAssets` have so I don't know if this example is correct, but it can be used to understand what I mean. Following this nomenclature, nowadays, a swap can be done with `L1Assets` and `SynthAssets` using Thornode quote swap endpoint, will this endpoint work with `TradeAssets` too? If the answer is yes, this feature can work with the `Asset` type, but, if the answer is no, this feature can not work with the `Asset` type and has to work only with `L1Assets` and `SynthAssets`. Due to the big impact this change has, it's important to define what can and what can not be done with this new type of asset.

Yes, you should be able to do swaps with trade assets, limited to TA<->TA and TA<->RUNE. 
A swap to and from a L1, from a TA is not allowed. They are a deposit or withdrawal.

PR #1066, opened 2024-03-05
1044 ClientLedger BTC

0xp3gasus:


Thorian1te (approved):
looks good, and I can see a way forward with asgardex now

PR #1079, opened 2024-03-17
Prepare examples to run in Codesandbox

hippocampus-web3:
I am working on improving the documentation of the library, attacking three points, in particular this PR is necessary to be able to work on point number three: 

1. Utilization of GitBook as the Central Documentation tool: We will use GitBook as our primary tool for documentation. It serves as the single access point to all other materials, streamlining the user experience and making it easier for developers to find what they need without navigating through multiple sources.

2. Generation of[ Interface Documentation with TypeDoc](https://docs.xchainjs.org/xchain-docs/): By leveraging TypeDoc, we automatically generate documentation from code comments. This approach ensures that our API documentation is always up-to-date and accurately reflects the codebase, reducing the manual effort required to keep documentation current.

3. Execution of Repository Examples in CodeSandbox: **To enhance the practical understanding of our library, we're executing existing examples in the repository directly within CodeSandbox. This integration allows users to interact with live examples, fostering a hands-on learning experience.**

Both the TypeDoc-generated documentation and CodeSandbox-ready examples are linked from our GitBook documentation. This integration ensures that users have a centralized point from which they can access all necessary materials to fully understand and effectively utilize our library.


Thorian1te (approved):
lgtm

PR #1224, opened 2024-07-29
Update stagenet derivation path for Radix

GiMa-Maya:
Don’t use radix testnet for Maya stagenet, but use radix mainnet


Thorian1te:
its broken bruv can you fix it and re push ?

PR #1227, opened 2024-08-02
1218 Thorchain Trade assets

0xp3gasus:


packages/xchain-aggregator/src/types.ts
```diff
@@ -14,9 +23,9 @@ type TxSubmitted = {
  * Fees
  */
 type Fees = {
-  asset: Asset | TokenAsset | SynthAsset // The asset for which fees are calculated
-  affiliateFee: CryptoAmount<Asset | TokenAsset | SynthAsset> // The affiliate fee amount
-  outboundFee: CryptoAmount<Asset | TokenAsset | SynthAsset> // The outbound fee amount
+  asset: Asset | TokenAsset | SynthAsset | TradeAsset // The asset for which fees are calculated
+  affiliateFee: CryptoAmount<Asset | TokenAsset | SynthAsset | TradeAsset> // The affiliate fee amount
+  outboundFee: CryptoAmount<Asset | TokenAsset | SynthAsset | TradeAsset> // The outbound fee amount
```
GrassRootsCrypto, line 28: Outbound fee won't be a trade asset. Trade Assets live within THORChain, even with Order books.

stoicdev6: there's any problem if i do this change?

Thorian1te (approved):
Had a Quick Look. 
Send it

packages/xchain-aggregator/src/types.ts
0xp3gasus: Looking at [documentation](https://thornode.ninerealms.com/thorchain/doc), in the QuoteFees type, the asset is the one used for all fees. If you are right, and outBoundFee is the native asset fee of the target fee. The bug is already there, we can create a new issue to fix

hippocampus-web3 (approved):

PR #1247, opened 2024-08-26
Fix bug EVM ledger signer

hippocampus-web3:
Fix bug with send and approve for LedgerSigner on EVM client


0xp3gasus:
Add the Changeset file with the change description

PR #1264, opened 2024-09-11
1258 MAYAChainAMM Radix support

0xp3gasus:


packages/xchain-mayachain-amm/src/mayachain-action.ts
```diff
@@ -87,47 +93,71 @@ export class MayachainAction {
...
+          methodName: 'user_deposit',
+          params: [
+            generateAddressParam(await wallet.getAddress(AssetXRD.chain)),
+            generateAddressParam(recipient),
+            generateBucketParam(0),
+            generateStringParam(memo),
+          ],
+        },
+      })
+      return {
```
juanfri-web3, line 119: Do you need the next `else` if you have this `return` here?

juanfri-web3 (approved):
Great!!

packages/xchain-mayachain-amm/src/mayachain-action.ts
0xp3gasus: Updated 💪🏻

juanfri-web3 (approved):

PR #1314, opened 2024-10-24
Thorchain.net as default explorer

0xp3gasus:


juanfri-web3 (approved):

Thorian1te:
just hold off on merging this. 
can we set our own default explorer?

Thorian1te (changes requested):
Im not sure I want this change. Can we have a different solution to this. I prefer runescan to thorchain.net

0xp3gasus:
> Im not sure I want this change. Can we have a different solution to this. I prefer runescan to thorchain.net

We can change the approach. We can use Runescan as the default explorer and add the possibility to let users to use a custom one

Thorian1te:
Yes, this would be useful indeed.

GrassRootsCrypto:
> Im not sure I want this change. Can we have a different solution to this. I prefer runescan to thorchain.net

I agree. Can we just pause this one for now please.

PR #1321, opened 2024-12-10
chore: remove redundant words in comment

crazeteam:


Thorian1te:
That is a generate file. There is no point making these changes. closing issue

PR #1347, opened 2025-02-16
feat: add approveRouter when mayachain-amm do-swap

hardworking-toptal-dev:
When we tried do-swap with mayachain-amm for real asset in EVM chains, it failes swapping.
So, I added approveRouter in mayachain-amm test file, it works well

### DIFF
--- a/examples/mayachain-amm/utils.ts
+++ b/examples/mayachain-amm/utils.ts
@@ -1,5 +1,7 @@
-import { QuoteSwap } from '@xchainjs/xchain-mayachain-query'
-import { assetToString } from '@xchainjs/xchain-util'
+import { QuoteSwap, CompatibleAsset } from '@xchainjs/xchain-mayachain-query'
+import { assetToString, Chain, TokenAsset, eqAsset, isSynthAsset } from '@xchainjs/xchain-util'
+import { AssetAETH } from '@xchainjs/xchain-arbitrum'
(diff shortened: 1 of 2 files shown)


Thorian1te (approved):

Thorian1te (dismissed):
nice work

examples/mayachain-amm/utils.ts
```diff
@@ -39,3 +41,23 @@ export const printQuoteSwap = (quoteSwap: QuoteSwap) => {
...
+export const isProtocolEVMChain = (chain: Chain): boolean => {
+  return [AssetETH.chain, AssetAETH.chain].includes(chain)
+}
+
+/**
+ * Check if asset is ERC20
+ * @param {Asset} asset to check
+ * @returns true if asset is ERC20, otherwise, false
+ */
+export const isProtocolERC20Asset = (asset: CompatibleAsset): asset is TokenAsset => {
```
Thorian1te, line 59: this function can be imported from the package 
`isProtocolERC20Asset` no need to rewrite it here

Thorian1te: okay let me fix this

Thorian1te: Btw the swap flow is meant to be like this 
Estimate swap => (in here it will check for approval via validateSwap() ) => asset needs approval ? call approve() : doSwap()

its all clearly laid out in the read me

PR #1370, opened 2025-04-03
add ankr api key param to provider

Thorian1te:


packages/xchain-avax/src/const.ts
```diff
@@ -13,8 +13,13 @@ export const AVAX_GAS_ASSET_DECIMAL = 18
 export const AVAXChain = 'AVAX' as const
 export const AssetAVAX: Asset = { chain: AVAXChain, symbol: 'AVAX', ticker: 'AVAX', type: AssetType.NATIVE }
 
+// Ankr api key
+const ankrApiKey = process.env.REACT_APP_ANKR_API_KEY
```
hippocampus-web3, line 17: I'm not a big fan of putting environment variables inside the library. I think it's better to pass the providers through the constructor, already configured with the application's keys. But I know that environment variables are used in many places. At least, can we avoid the REACT_APP prefix so it doesn't cause confusion or conflicts with certain frameworks or bundles in other environments?

Thorian1te: we can :)

hippocampus-web3 (approved):

PR #1373, opened 2025-04-12
Zcash client

hippocampus-web3:
Add a new client for Zcash

⚠️ Only supports T-ADDRESSES ⚠️

# Notes
The initial node provider is NOWNodes, though additional providers may be integrated in the future based on demand.

Address construction, signing, and validation logic is handled via a package developed by members of the Zcash community. For now, the NPM package is published under my handle, but the original source code can be found here:
👉 https://gitlab.com/mayachain/chains/zcash/-/tree/main/js?ref_type=heads
I’ve only made some modifications to adapt it to the Xchain architecture.

# Future Work

- Ledger support

- PrepareTx support: The prepareTx function, which generates an unsigned hexadecimal transaction, is currently unimplemented due to its complexity and limited usage. If needed in the future, we can explore implementation options.


Thorian1te (approved):
lgtm

PR #1392, opened 2025-05-15
yarn dedupe

dynst:


Thorian1te (approved):
lgtm

hippocampus-web3:
@dynst this looks like breaks the build command

dynst:
Running `yarn run clean` somehow is not enough to guarantee the build actually starts over again from a clean state, because I still couldn't reproduce the failure in `xchain-kujira` that CI revealed. It just kept inexplicably succeeding.

```
@xchainjs/xchain-kujira:build: cache miss, executing a02c0204b66cde91
@xchainjs/xchain-kujira:build:
@xchainjs/xchain-kujira:build: src/index.ts → lib/index.js, lib/index.esm.js...
@xchainjs/xchain-kujira:build: created lib/index.js, lib/index.esm.js in 5.1s
```

No matter how deleted the output dir `xchain-kujira/lib/` was, it was still succeeding - I can only guess the rollup TypeScript plugin must cache the validation result based on a hash of the input file or something. `touch packages/xchain-kujira/src/const.ts` doesn't force it to re-check. The file's contents have to be changed before it'll notice something wrong.

dynst:
Now that I finally found a way to reproduce the CI errors locally, there are two culprit dependencies: `bignumber.js` and `bip174` happen to be pinned at versions that have allowed invalid TypeScript code to be added to the repo. Updating them makes validation fail because suddenly it's much stricter.

Didn't find a way to fix `bip174` fallout easily so just leaving that one out for now. Updated the code with new annotations where BigNumber is involved (via BaseAmount) and did `yarn run update-packages`.

Can a setting be changed to start letting anyone's PR trigger a CI run in the future?

dynst:
Anything blocking this now?

PR #1397, opened 2025-05-19
remove coininfo

xcaptain:
coininfo depends on `safe-buffer`, which is not loadable by vite, a lots of new frontend framework use vite to bundle dependencies. This PR make it easier for people to integrate xchainjs into a frontend project.

### DIFF
--- a/packages/xchain-util/package.json
+++ b/packages/xchain-util/package.json
@@ -31,7 +31,9 @@
     "lint": "eslint \"{src,__tests__}/**/*.ts\" --fix --max-warnings 0"
   },
   "dependencies": {
-    "bignumber.js": "9.0.0"
+    "bignumber.js": "9.0.0",
+    "jest": "^29.7.0",
+    "test": "^3.3.0"
   },
   "publishConfig": {
     "access": "public",
--- a/packages/xchain-util/src/coininfo.ts
+++ b/packages/xchain-util/src/coininfo.ts
@@ -1,0 +1,75 @@
+import * as bchData from './coins/bch'
+import * as dogeData from './coins/doge'
+import * as dashData from './coins/dash'
+import * as ltcData from './coins/ltc'
+
+
+interface CoinVersion {
+    bip32: {
+        public: number
+        private: number
+    }
+    bip44?: number
+    private: number
+    public: number
+    scripthash: number
+}
+
+interface CoinConfig {
+    name: string
+    unit: string
+    hashGenesisBlock?: string
+    port?: number
+    protocol?: {
+        magic: number
+    }
+    seedsDns?: string[]
+    versions: CoinVersion
+    messagePrefix?: string
+    bech32?: string
+    testnet?: boolean
+}
+
+interface CoinData {
+    main: CoinConfig
+    test?: CoinConfig
+    regtest?: CoinConfig
+    simnet?: CoinConfig
+}
+
+function toBitcoinJSInner(coinConfig: CoinConfig) {
+    return {
+        messagePrefix: coinConfig.messagePrefix || '\x19' + coinConfig.name + ' Signed Message:\n',
+        bech32: coinConfig.bech32,
+        bip32: {
+            public: (coinConfig.versions.bip32 || {}).public,
+            private: (coinConfig.versions.bip32 || {}).private
+        },
+        pubKeyHash: coinConfig.versions.public,
+        scriptHash: coinConfig.versions.scripthash,
+        wif: coinConfig.versions.private,
+        dustThreshold: null
+    }
+}
+
+const coinConfigs: Record<string, CoinData> = {
+    bitcoincash: bchData,
+    dogecoin: dogeData,
+
+    dash: dashData,
+    litecoin: ltcData,
+}
+
+export function toBitcoinJS(chain: 'bitcoincash' | 'dogecoin' | 'dash' | 'litecoin', network: 'main' | 'test') {
+    const coinData = coinConfigs[chain]
+    if (!coinData) {
+        throw new Error(`Coin data for ${chain} not found`)
+    }
+
+    const config = network === 'main' ? coinData.main : coinData.test
+    if (!config) {
+        throw new Error(`Network ${network} not found for ${chain}`)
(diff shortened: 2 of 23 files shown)


xcaptain:
@hippocampus-web3 can you take a look, do you think it's a good idea to remove coininfo?

xcaptain:
Still think it's good to delete coininfo dependency, coininfo 5.1.0 is too old(6 years ago), which depends on safe-buffer 5.1.1 (8 years ago)
image

these old dependencies are not compatitable with modern web app bundlers like vite.

johnnynanjiang:
Hi @hippocampus-web3 , can you please help with this, as well as advice the overall dev process for xchainjs, as we at Vastlink are more than happy to contribute.

packages/xchain-util/package.json
```diff
@@ -31,7 +31,9 @@
     "lint": "eslint \"{src,__tests__}/**/*.ts\" --fix --max-warnings 0"
   },
   "dependencies": {
-    "bignumber.js": "9.0.0"
+    "bignumber.js": "9.0.0",
+    "jest": "^29.7.0",
+    "test": "^3.3.0"
```
hippocampus-web3, line 36: I don't think these dependencies belong here.

packages/xchain-util/src/coininfo.ts
```diff
@@ -0,0 +1,75 @@
+import * as bchData from './coins/bch'
+import * as dogeData from './coins/doge'
+import * as dashData from './coins/dash'
+import * as ltcData from './coins/ltc'
+
+
+interface CoinVersion {
```
hippocampus-web3, line 7: I think we should put this in the `xchainjs-utxo` instead of `xchainjs-utils` package since it only applies to UTXO chains

hippocampus-web3 (changes requested):

hippocampus-web3:
Yes, I think it's a good idea to get rid of this dependencym thanks for PR @xcaptain 

To contribute @johnnynanjiang , you can read the [CONTRIBUTING.md](https://github.com/xchainjs/xchainjs-lib/blob/master/CONTRIBUTING.md.)

But here’s a quick summary:

1. You open a PR.

2. It gets reviewed and the CI is run to make sure everything works.

3. Run yarn update to generate a changelog with the new changes.

4. It gets merged into master.

packages/xchain-util/package.json
xcaptain: You are right, I added it by mistake, will change it.

packages/xchain-util/src/coininfo.ts
hippocampus-web3: The provider package only contain integrations with external data providers. I think the correct place for this is xchain-utxo. You can copy the test configuration from any other package.

xcaptain:
coin definitions are copied from the coininfo repository. tests are generated from this script:

```js
import coininfo from "coininfo";

const coin = coininfo.dash.test.toBitcoinJS(); // replace with bitcoincash, dash, litecoin, dogecoin
console.log('coin', coin)
```

packages/xchain-util/src/coininfo.ts
xcaptain: moved to `xchain-utxo` and added a test file.

packages/xchain-utxo/src/coininfo.ts
```diff
@@ -0,0 +1,74 @@
...
+    return {
+        messagePrefix: coinConfig.messagePrefix || '\x19' + coinConfig.name + ' Signed Message:\n',
+        bech32: coinConfig.bech32,
+        bip32: {
+            public: (coinConfig.versions.bip32 || {}).public,
+            private: (coinConfig.versions.bip32 || {}).private
+        },
+        pubKeyHash: coinConfig.versions.public,
+        scriptHash: coinConfig.versions.scripthash,
+        wif: coinConfig.versions.private,
```
hippocampus-web3, line 50: Why remove dustThreshold ? 

https://github.com/cryptocoinjs/coininfo/blob/e744f1caf1529e7feccd98aac889214dc2bd2cc7/lib/coininfo.js#L93

hippocampus-web3 (changes requested):

hippocampus-web3:
You can apply cast like this, to avoid build errores (they are present on doge- dash and litecoin client)

```ts
    case Network.Stagenet:
      return toBitcoinJS('dash', 'main') as Dash.Network // Convert Dash mainnet to BitcoinJS format
    case Network.Testnet:
      return toBitcoinJS('dash', 'test') as Dash.Network 
```

Also execute locally `yarn lint` so fix format 🙏 @xcaptain

packages/xchain-utxo/src/coininfo.ts
xcaptain: I try to keep consistant with `bitcoinjs-lib`, but nevermind, I added it back.

xcaptain:
`yarn lint`  applied and `yarn build` passed, please take a look again.

xcaptain:
CI failed, should be related to https://github.com/xchainjs/xchainjs-lib/issues/1402

because in `bitcoinjs-lib` v5.2.0, bech32 is a string type.

https://github.com/bitcoinjs/bitcoinjs-lib/blob/v5.2.0/types/networks.d.ts#L3

not a big deal, we can merge this first.

hippocampus-web3:
We can't merge the broken CI. It's the same issue as with Dash, you can fix it with a cast in the clients that are failing.
You need to update it in Doge and Litecoin. Make sure the build passes locally. Once it passes, you must run `yarn update-packages` as the final step prior to merge and describe the nature of your changes, so that the CI can automatically bump the package versions on merge. Otherwise, the changes won’t be published. [CONTRIBUTING.md](https://github.com/xchainjs/xchainjs-lib/blob/master/CONTRIBUTING.md.)

hippocampus-web3:
lgtm ! Take a look if you can @Thorian1te :pray:

Thorian1te (approved):
lgtm

hippocampus-web3 (approved):

PR #1410, opened 2025-06-02
use "exports" field in package.json

dynst:
https://nodejs.org/api/packages.html#exports

Without this field, Node.js import logic believes it's just a CJS package - node doesn't recognize the old non-standard `"module"` field (while transpilers like webpack do).

### DIFF
--- a/.changeset/rude-dancers-buy.md
+++ b/.changeset/rude-dancers-buy.md
@@ -1,0 +1,42 @@
+---
+'@xchainjs/xchain-mayamidgard-query': minor
+'@xchainjs/xchain-mayachain-query': minor
+'@xchainjs/xchain-thorchain-query': minor
+'@xchainjs/xchain-utxo-providers': minor
+'@xchainjs/xchain-evm-providers': minor
+'@xchainjs/xchain-mayachain-amm': minor
+'@xchainjs/xchain-midgard-query': minor
+'@xchainjs/xchain-thorchain-amm': minor
+'@xchainjs/xchain-bitcoincash': minor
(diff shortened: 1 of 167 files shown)


Thorian1te (dismissed):
fantastic

dynst:
Never mind. The built ESM code in `xchain-cosmos`/kujira/thorchain/mayachain (and probably anything that depends on them) just doesn't work.

It's probably never worked, and has just never been tested before. It fails trying to import from https://github.com/cosmos/cosmjs-types/issues/93

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module 'xchainjs-lib/node_modules/cosmjs-types/cosmos/tx/v1beta1/tx' imported from xchainjs-lib/packages/xchain-cosmos/lib/index.esm.js
Did you mean to import cosmjs-types/cosmos/tx/v1beta1/tx.js?
```

https://github.com/xchainjs/xchainjs-lib/blob/842920f6cb0581e221d3fb4b381c7eaf5890bd9c/packages/xchain-cosmos/src/client.ts#L15

It looks like there's no actual ESM-compatible way to access that type, because it's not even exported from the top level [index.ts](https://github.com/cosmos/cosmjs-types/blob/main/src/index.ts).

CosmJs packages only very lightly maintained by Confio these days. https://github.com/cosmos/cosmjs/issues/1649#issuecomment-2909668587

Thorian1te:
Closing stale draft: `exports` field work is not moving forward and has been open since 2025. Reopen or open a fresh PR if/when this is picked up again.

PR #1411, opened 2025-06-03
fix faulty tsconfig outDir

dynst:
Rollup handles the code output, typescript only does the declarations.


Thorian1te (commented):
Nice, keeps the project cleaner by isolating type declarations in one directory

dynst:
This is a prerequisite for #1447

PR #1420, opened 2025-06-12
feat: add xchain-ripple client

mrtommyit:
## Features

- Added xchain-ripple package
- Abstract `Client` to be used for wallets
- Implemented `ClientKeystore` for keystore wallet


packages/xchain-ripple/src/client.ts
```diff
@@ -0,0 +1,386 @@
...
+   * @returns {Fees} The average, fast, and fastest fees.
+   * @throws {"Params need to be passed"} Thrown if parameters are not provided.
+   */
+  public async getFees(params?: FeeEstimateOptions & { amount: BaseAmount; sender: Address }): Promise<Fees> {
+    if (!params) {
+      throw Error('Params not provided')
+    }
+    const xrplClient = await this.getXrplClient()
+    const feeResponse = await xrplClient.request({ command: 'fee' })
+    const fee = feeResponse.result.drops.open_ledger_fee
```
Thorian1te: Q. if there is no feeResponse, should this return fee error or default fees

mrtommyit: Ok, I've set the default fee as `10` drops if failed to get fee from xrpl.

mrtommyit:
@Thorian1te , fixed yarn.lock file now.

packages/xchain-ripple/src/client.ts
```diff
@@ -0,0 +1,380 @@
...
+    throw new Error('not supported')
+  }
+
+  async getFeesWithRates(): Promise<FeesWithRates> {
+    throw Error('not supported')
+  }
+
+  async getFeeRates(): Promise<FeeRate> {
+    throw Error('not supported')
+  }
```
hippocampus-web3: What is the reason for discontinuing support for these features?

mrtommyit: I've implemented `broadcastTx` and added e2e test.
image

- `getFeesWithRates` and `getFeeRates` are not supported because Ripple uses the flat fee.

- `prepareTx` is not supported because client is using `xrpl` standard and it requires `Payment` type to be signed from any XRP wallets (ledger, xaman, etc). Instead, `prepareTxForXrpl` is implemented.

hippocampus-web3: Understood, so in the `getFees` function, the multipliers should be removed? https://github.com/xchainjs/xchainjs-lib/pull/1420/files#diff-ef07ff9a12bdcadc5a1d0e96c4ae80caf4d5b8c858e3f8fbd1fb20ac56a73680R188

mrtommyit: Exactly! I removed fee multipliers now.

hippocampus-web3 (approved):

PR #1444, opened 2025-06-26
Fix approve max amount

hippocampus-web3:
FIx for https://github.com/asgardex/asgardex-desktop/pull/790

### DIFF
--- a/.changeset/true-mice-sing.md
+++ b/.changeset/true-mice-sing.md
@@ -1,0 +1,5 @@
+---
+'@xchainjs/xchain-evm': patch
+---
+
+Fix function prepareApprove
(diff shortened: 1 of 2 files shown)


dynst:
Do these lines in `approve` need changing too?

https://github.com/xchainjs/xchainjs-lib/blob/8c583938ff03f9d950554c13361953ae0f57597d/packages/xchain-evm/src/clients/client.ts#L854-L855

hippocampus-web3:
At some point, we could consider addressing the change from bignumber.js, but it's so deeply integrated into the library that I'm not sure the time spent would be worth the benefit gained

PR #1450, opened 2025-06-30
update for native thor.rujis support

Thorian1te:
balances were being returned like so.
Screenshot 2025-06-30 at 4 29 59 PM

 since `X/RUJI` is unique it needs to be handled like so. Asset should be `THOR.RUJI` to work with thorchain's deposit & pool functions.


Thorian1te:
now we have 
Screenshot 2025-06-30 at 4 30 08 PM

hippocampus-web3 (approved):

PR #1457, opened 2025-07-07
update to typescript 5.5

dynst:


hippocampus-web3 (changes requested):
All examples still depend on Typescript 5.0.4. Please update that as well.

dynst:
> All examples still depend on Typescript 5.0.4.

They depend on `^5.0.4` so the caret actually means they're also now also using 5.5.

```
"typescript@npm:^5.0.4, typescript@npm:^5.5":
  version: 5.5.4
```

hippocampus-web3:
And just out of curiosity, why did we define 5.5? And not the most recent version?

hippocampus-web3:
I don't quite see the point of updating only up to 5.5 in the package.json files when the caret allows us to go up to 5.8 — it's confusing. Let's align all the package.json files to 5.8 and use the caret so they can still be upgraded, but since we're already updating, let's set everything to the version we're actually using.

dynst:
Sure. Rebased against the new `examples/loans/` removal.

hippocampus-web3 (approved):

PR #1458, opened 2025-07-07
update @types/node to 22.7.5

dynst:


hippocampus-web3:
I think "@types/node": "*" it's too agresive. Please use a specific version

hippocampus-web3 (approved):

Issue #1449, opened 2025-06-30
completing the noble cryptography migration


dynst:
- [x] The EVM chains migrated with the update to ethers v6 and most packages dropped `bip32` in #1424
- [ ] replace `@bitcoin-js/tiny-secp256k1-asmjs` and `ecpair`
- [ ] replace `bech32` with `@scure/base`
- [ ] xchain-radix currently depends on `bip32` and `ed25519-hd-key` instead of `@scure/bip32` and `micro-key-producer`
- [ ] crypto-js, bip39, and the custom fixed-up [fork](https://github.com/xchainjs/foundry-primitives-js) of foundry-primitives are replaceable

hippocampus-web3:
This `replace @bitcoin-js/tiny-secp256k1-asmjs and ecpair` will be delayed until later because it requires us to migrate from bitcoinjs-lib, and we do not want to do that at this time.

PR #1462, opened 2025-07-12
Remove bech32 from first level dependency

hippocampus-web3:
Remove bech32 as the first level dependency. Part of the work done for the issue 

https://github.com/xchainjs/xchainjs-lib/issues/1449


packages/xchain-crypto/package.json
```diff
@@ -64,4 +55,4 @@
     "access": "public",
     "directory": "release/package"
   }
-}
+}
```
dynst: The trailing newline got stripped by accident, can the newline be put back?

Otherwise lgtm

dynst (dismissed):

dynst (approved):
removing the defunct `getLoanQuoteOpen` / `getLoanQuoteClose` examples too is great

PR #1474, opened 2025-07-16
Fix package ecosystem dependabot

hippocampus-web3:


dynst:
The reason for using `npm` now is because `yarn` v4 isn't supported yet, right? https://github.com/dependabot/dependabot-core/issues/8610

So when they add support it would make sense to switch back then.

PR #1480, opened 2025-07-17
import fixes

dynst:
Importing from CJS-only packages, namespace imports only work if named imports work, and named imports only work if the static analysis heuristic that Node.js uses can find them, which it can't, in the case of `bitcore-lib-cash` and `@dashevo/dashcore-lib`. (Static analysis can find them for the more simply-written `bchaddr`.)

https://nodejs.org/api/esm.html#commonjs-namespaces

```
SyntaxError: Named export 'Network' not found. The requested module 'bitcore-lib-cash' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'bitcore-lib-cash';
const { Network } = pkg;
```

`bignumber.js` had busted TypeScript type definitions (for ESM) until v9.1.2. v9.1.0 was the first version with any ESM support at all.

Apply a workaround for https://github.com/protobufjs/protobuf.js/issues/2047 in generated code.

Prerequisites for #1410

(I'm guessing thorchain and mayachain protobuf code hasn't been regenerated in a while, because a lot of new code suddenly appeared in `MsgCompiled` re-running the script.)

## Summary by CodeRabbit

- Bug Fixes
  - Resolved ESM compatibility issues across multiple chain packages and examples, ensuring smooth usage in Node ESM and modern bundlers.
  - Corrected import paths and default imports where required; updated generated protobuf imports for ESM.

- Chores
  - Bumped bignumber.js to 9.1.2 across the repo.
  - Applied patch version updates to affected packages.

- Tests
  - Updated test imports to align with ESM-compatible setups.


hippocampus-web3:
@coderabbitai full review

Thorian1te (approved):

PR #1403, opened 2025-05-21
Update dependencies

hippocampus-web3:

PR #1481, opened 2025-07-17
use semver caret ^ dependencies for ledger packages

dynst:
Applications like asgardex shouldn't be locked to one particular version of ledger packages or axios.

### DIFF
--- a/.changeset/rich-ends-find.md
+++ b/.changeset/rich-ends-find.md
@@ -1,0 +1,25 @@
+---
+'@xchainjs/xchain-mayamidgard-query': patch
+'@xchainjs/xchain-utxo-providers': patch
+'@xchainjs/xchain-evm-providers': patch
+'@xchainjs/xchain-mayachain-amm': patch
+'@xchainjs/xchain-midgard-query': patch
+'@xchainjs/xchain-thorchain-amm': patch
+'@xchainjs/xchain-bitcoincash': patch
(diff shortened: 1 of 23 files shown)


dynst:
>  we ran into this just a few weeks ago.

I agree that's an issue, and I think the lack of caret ^ semver versioning was the ultimate *cause* of that issue. https://github.com/xchainjs/xchainjs-lib/pull/1403#issuecomment-2968332337

The status quo is, the version of the first-level Ledger dependency is pinned, but the dependencies underneath it *aren't*. That's what that dedupe did, it updated a second-level or lower dependency, because Ledger uses a caret ^ to depend on it, so it was allowed to be updated.

Pinning or not pinning the first-level dependency can't stop second-level dependencies from being updated. Given that as a library, xchainjs can't stop that - an application can set `"resolutions"` if it really wants to pin lower-level dependencies to specific versions - the best thing to do is allow first-level dependencies to be updated at the same time second-level dependencies get updated, so both are the latest version (at the time of install).

Installing an old first-level dependency with a brand new second-level dependency is the combination that's likely to cause problems, and the current versioning constraint is likely to cause that again. If second-level dependencies are unconstrained, first-level dependencies have to be, too.

hippocampus-web3:
Okay, let's try it!

hippocampus-web3 (approved):

PR #1526, opened 2025-08-27
phase 1 utxo improvements

Thorian1te:


packages/xchain-utxo/src/utxo-selector.ts
```diff
@@ -0,0 +1,512 @@
...
+      inputCount * UtxoSelector.BYTES_PER_INPUT +
+      outputCount * UtxoSelector.BYTES_PER_OUTPUT
+    return Math.ceil(txSize * feeRate)
+  }
+}
+
+/**
+ * Branch and Bound strategy - optimal for minimizing fees and change
+ */
+export class BranchAndBoundStrategy implements UtxoSelectionStrategy {
```
hippocampus-web3: Split one strategy per file to improve readability

packages/xchain-utxo/src/validators.ts
```diff
@@ -0,0 +1,407 @@
...
+        utxoCount: utxos.length,
+        validationErrors: errors,
+      })
+    }
+  }
+
+  /**
+   * Validate address format for specific network
+   */
+  static validateAddressFormat(address: string, network: Network, chain: string, allowedFormats?: string[]): void {
```
hippocampus-web3: We already have an address validation function per client. Should we use them ?. Maybe address validation should be at specifica package level (bitcoin, litecoin, etc..) because it's different validation per chain.

packages/xchain-utxo/src/validators.ts
```diff
@@ -0,0 +1,407 @@
...
+    if (!address) {
+      throw UtxoError.invalidAddress('', network)
+    }
+
+    if (typeof address !== 'string') {
+      throw UtxoError.validationError('Address must be a string')
+    }
+
+    // Basic length check
+    if (address.length < 10 || address.length > 100) {
```
hippocampus-web3, line 227: This seems like an arbitrary check that does not provide much value

packages/xchain-utxo/src/validators.ts
```diff
@@ -0,0 +1,407 @@
...
+
+  /**
+   * Validate address format for specific network
+   */
+  static validateAddressFormat(address: string, network: Network, chain: string, allowedFormats?: string[]): void {
+    if (!address) {
+      throw UtxoError.invalidAddress('', network)
+    }
+
+    if (typeof address !== 'string') {
```
hippocampus-web3: Should we check this when we are using Typescript ?

packages/xchain-bitcoin/src/client.ts
```diff
@@ -230,11 +471,204 @@ abstract class Client extends UTXOClient {
...
+      throw UtxoError.fromUnknown(error, 'prepareTxEnhanced')
+    }
+  }
+
   /**
    * Prepare transfer.
    *
    * @param {TxParams&Address&FeeRate&boolean} params The transfer options.
    * @returns {PreparedTx} The raw unsigned transaction.
+   * @deprecated Use prepareTxEnhanced for better performance and validation
```
hippocampus-web3: The new method respects the signature of the old one, so why not release a major version, remove the old method, and thus clean up the codebase and reduce dependencies, instead of just marking it as deprecated?

packages/xchain-bitcoin/src/client.ts
```diff
@@ -230,11 +471,204 @@ abstract class Client extends UTXOClient {
...
+      }
+
+      UtxoTransactionValidator.validateAddressFormat(recipient, this.network, 'BTC')
+      UtxoTransactionValidator.validateAddressFormat(sender, this.network, 'BTC')
+
+      if (memo) {
+        // Only validate memo for send max (skip amount validation)
+        if (typeof memo !== 'string') {
+          throw UtxoError.validationError('Memo must be a string')
+        }
```
hippocampus-web3: Is this check neccessary  in Typescript ?

packages/xchain-utxo/src/validators.ts
```diff
@@ -0,0 +1,407 @@
...
+        const amount = params.amount.amount()
+
+        if (amount.lte(0)) {
+          errors.push('Amount must be greater than zero')
+        }
+
+        // Check for reasonable maximum (21M BTC equivalent in satoshis)
+        if (amount.gt('2100000000000000')) {
+          errors.push('Amount exceeds maximum possible value')
+        }
```
hippocampus-web3: I don't think this validation makes much sense, at least at utxo package level

packages/xchain-utxo/src/validators.ts
```diff
@@ -0,0 +1,407 @@
...
+      sender?: Address
+      feeRate?: number
+    },
+  ): void {
+    const errors: string[] = []
+
+    // Address validation
+    if (!params.recipient?.trim()) {
+      errors.push('Recipient address is required')
+    } else if (params.recipient.length > 100) {
```
hippocampus-web3: Why `length > 100` ?

packages/xchain-utxo/src/validators.ts
```diff
@@ -0,0 +1,407 @@
...
+          errors.push('Amount must be greater than zero')
+        }
+
+        // Check for reasonable maximum (21M BTC equivalent in satoshis)
+        if (amount.gt('2100000000000000')) {
+          errors.push('Amount exceeds maximum possible value')
+        }
+
+        // Check for dust amount (546 satoshis is standard dust threshold)
+        if (amount.lt(546)) {
```
hippocampus-web3: I think dust threshole it's defiend as constant in other place. Maybe reuse constant instead of use magic number. Also wondering. Dush threshold can be equal for all UTXO changes ?

packages/xchain-utxo/src/validators.ts
```diff
@@ -0,0 +1,407 @@
...
+          errors.push('Amount is below dust threshold (546 satoshis)')
+        }
+
+        // Validate decimal precision
+        const decimals = params.amount.decimal
+        if (decimals < 0 || decimals > 18) {
```
hippocampus-web3: At the moment, all integrated UTXOs have 8 decimal places, why between 0 and 18 ?

packages/xchain-utxo/src/validators.ts
```diff
@@ -0,0 +1,407 @@
...
+    // Fee rate validation
+    if (params.feeRate !== undefined) {
+      if (!Number.isFinite(params.feeRate)) {
+        errors.push('Fee rate must be a finite number')
+      } else if (params.feeRate < 1) {
+        errors.push('Fee rate must be at least 1 sat/byte')
+      } else if (params.feeRate > 1000000) {
+        errors.push('Fee rate is unreasonably high (max 1,000,000 sat/byte)')
+      }
+    }
```
hippocampus-web3, line 87: Would it make sense to use the fee bounds here instead hardcoded values ?

packages/xchain-utxo/src/validators.ts
```diff
@@ -0,0 +1,407 @@
...
+
+    // Memo validation
+    if (params.memo !== undefined) {
+      if (typeof params.memo !== 'string') {
+        errors.push('Memo must be a string')
+      } else {
```
hippocampus-web3: Is the string check necessary? I also think this validation is being done several times

hippocampus-web3 (changes requested):

packages/xchain-bitcoin/src/client.ts
Thorian1te: good point

PR #1528, opened 2025-08-27
update & pin elliptic

Thorian1te:
## Summary by CodeRabbit

* **Chores**
  * Updated dependency resolutions to include a specific version for a cryptography library, ensuring consistent installs across environments. Existing resolution for a multipart/form-data library remains unchanged. This is a maintenance-only update with no user-facing changes; functionality and UI are unaffected. No action is required from users. App performance and features remain the same.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,8 @@
     "micromatch": "4.0.8",
     "word-wrap": "1.2.4",
     "undici": "5.29.0",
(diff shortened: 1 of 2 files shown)


Naq302 (approved):
Dup of https://github.com/xchainjs/xchainjs-lib/pull/1522 - good work

PR #1532, opened 2025-09-04
Feat/evm performance optimizations

Thorian1te:
## Summary by CodeRabbit

- New Features
  - Added MAYAChain gas price estimation support.

- Performance
  - In-memory caching for contract and numeric objects to reduce allocations and speed repeated operations.
  - Parallelized provider queries and gas/limit estimations for faster responses.

- Reliability
  - Improved round‑robin provider handling with clearer errors when no providers are available.

- Tests
  - Added tests verifying per-provider and per-address contract caching behavior.

### DIFF
--- a/.changeset/flat-melons-bathe.md
+++ b/.changeset/flat-melons-bathe.md
@@ -1,0 +1,5 @@
+---
+'@xchainjs/xchain-evm': minor
+---
+
+Added caching for better performance


packages/xchain-evm/src/cache.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Get a cached BigNumber instance or create a new one
+ * Only accepts string or bigint to preserve precision
+ */
+export function getCachedBigNumber(value: string | bigint): BigNumber {
+  const stringValue = typeof value === 'bigint' ? value.toString() : value
+  if (!bigNumberCache.has(stringValue)) {
+    bigNumberCache.set(stringValue, new BigNumber(stringValue))
+  }
+  return bigNumberCache.get(stringValue)!
+}
```
hippocampus-web3, line 44: What is this cache trying to optimize? It seems that it only saves the BigNumber instance. Is this really a costly process?

packages/xchain-evm/src/clients/client.ts
```diff
@@ -77,6 +78,33 @@ export type EVMClientParams = XChainClientParams & {
   signer?: ISigner
 }
 
+/**
+ * Helper function to race promises and return the first successful result
+ * Uses Promise.race with proper error handling as fallback for Promise.any
+ */
+async function promiseAny<T>(promises: Promise<T>[]): Promise<T> {
+  // Use Promise.race to get the first resolved promise
```
hippocampus-web3, line 86: Would it make sense to use promise.any ?

PR #1666, opened 2026-03-23
Feature/blockbook

sourovafrin:
## Summary by CodeRabbit

* **New Features**
  * Added Blockbook-based data providers for Bitcoin, Bitcoin Cash, Dogecoin, and Litecoin to fetch balances, UTXOs, transaction history/details, broadcast raw transactions, and estimate fees.
  * Providers include optional API-key support, address normalization, and paginated transaction queries for more reliable remote lookups.
* **Chores**
  * Minor-version bumps applied to the affected packages.

### DIFF
--- a/.changeset/icy-apples-warn.md
+++ b/.changeset/icy-apples-warn.md
@@ -1,0 +1,9 @@
+---
+'@xchainjs/xchain-utxo-providers': minor
+'@xchainjs/xchain-bitcoincash': minor
+'@xchainjs/xchain-litecoin': minor
+'@xchainjs/xchain-bitcoin': minor
+'@xchainjs/xchain-doge': minor
+---
+
+Added Blockbook data providers for Bitcoin, Bitcoin Cash, Dogecoin, and Litecoin. Implemented BlockbookProvider for each asset, including API methods for transactions and balance retrieval
--- a/packages/xchain-bitcoin/src/client.ts
+++ b/packages/xchain-bitcoin/src/client.ts
@@ -18,6 +18,7 @@ import {
   BTCChain,
   BTC_DECIMAL,
   BitgoProviders,
+  BlockbookDataProviders,
   BlockcypherDataProviders,
   LOWER_FEE_BOUND,
   MIN_TX_FEE,
@@ -32,7 +33,7 @@ export const defaultBTCParams: UtxoClientParams = {
   network: Network.Mainnet,
   phrase: '',
   explorerProviders: blockstreamExplorerProviders,
-  dataProviders: [BitgoProviders, BlockcypherDataProviders],
+  dataProviders: [BitgoProviders, BlockcypherDataProviders, BlockbookDataProviders],
   rootDerivationPaths: {
     [Network.Mainnet]: `m/84'/0'/0'/0/`, // Not BIP44 compliant but compatible with pre-HD wallets
     [Network.Testnet]: `m/84'/1'/0'/0/`,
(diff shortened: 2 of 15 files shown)


Thorian1te (dismissed):

Thorian1te (commented):
## Code Review: Blockbook Data Providers

### Critical Issues

**1. `apiKey` added to `UtxoOnlineDataProvider` interface** (`packages/xchain-utxo-providers/src/types/index.ts`)

This is an implementation detail, not a contract. `NownodesProvider` already handles `apiKey` at the class level without polluting the interface. Providers like `BitgoProvider`/`HaskoinProvider` don't use API keys. Remove from the interface, keep on the class.

**2. `getFeeRate?(feeOption)` on the interface is premature** (same file)

Adding an optional method to a shared interface for one provider's optimization should be a separate PR if needed at all. The `roundRobinGetFeeRate` fallback already handles calling `getFeeRates()` when this method is absent.

**3. BTC Blockbook providers defined but never wired in**

`BlockbookDataProviders` is exported from `xchain-bitcoin/src/const.ts` but never added to `defaultBTCParams.dataProviders` in `client.ts`. BCH, DOGE, and LTC all get wired in — BTC does not. Bug or intentional?

**4. Changeset missing `@xchainjs/xchain-utxo`**

The PR adds `roundRobinGetFeeRate` and `getFeeRate` to the UTXO base client (`packages/xchain-utxo/src/client.ts`), which is new public API surface. Needs a version bump in the changeset.

---

### High Priority

**5. 500ms sleep per UTXO in `mapUTXOs` will cause severe performance issues** (`blockbook-data-provider.ts`)

20 unique UTXOs = 10 seconds of artificial delay. 100 UTXOs = 50 seconds. The existing `NownodesProvider` doesn't need to fetch individual transactions for each UTXO. If rate limiting is truly needed, use 50-100ms or batch requests. Consider whether `witnessUtxo` is even needed when `txHex` is already available.

**6. Hardcoded `nativeswap.io` URLs with no documentation**

BCH/DOGE/LTC all route through `{chain}.nativeswap.io`. Who operates this? What are the uptime/rate-limit guarantees? Routing user addresses through an unknown third party is a privacy concern. BTC uses NowNodes (known provider with API key auth), but the others don't.

**7. `process.env.NOWNODES_API_KEY || ''`** (`xchain-bitcoin/src/const.ts`)

Evaluated at module load time. An empty string API key sent to NowNodes may cause 401 errors rather than graceful degradation. Consistent with existing patterns but worth noting.

---

### Medium Priority

**8. Pagination fetches excess data** (`blockbook-data-provider.ts` — `getRawTransactions`)

With offset=990, limit=10, this fetches 1000 txs to return 10. Blockbook's API supports a `page` parameter for proper server-side pagination.

**9. No validation on fee estimate** (`blockbook-api.ts` — `getFeeEstimate`)

`parseFloat(dto.result)` can return `NaN` or negative values (Blockbook returns `"-1"` when it can't estimate). This silently propagates bad fee rates. Suggest:
```typescript
const btcPerKb = parseFloat(dto.result)
if (isNaN(btcPerKb) || btcPerKb <= 0) {
  throw new Error(`Invalid fee estimate from Blockbook: ${dto.result}`)
}
```

**10. No tests**

Zero test coverage for any new code. Existing providers (Haskoin, Blockcypher, etc.) have test files. At minimum needs mocked unit tests for the API layer and mapping logic.

---

### Low Priority

**11. `encodeAddressPathSegment` over-engineered** — Standard BTC/LTC/DOGE addresses never contain colons. BCH uses `toCashAddress` normalization before reaching this code.

**12. Inconsistent vin/vout filtering in `mapTransactionToTx`** — `vin` only checks `addresses?.length > 0`, but `vout` also checks `isAddress`. Could include coinbase inputs.

**13. No provider name in error messages** — Makes debugging harder in round-robin setups. Other providers include their name in errors.

Thorian1te (approved):

Issue #1715, opened 2026-06-25
xchain-crypto 1.0.7: crypto.timingSafeEqual breaks keystore decrypt in browser/bundler environments


underthesun49:
### Summary

`@xchainjs/xchain-crypto` **1.0.7** changed `decryptFromKeystore` to verify the keystore MAC with Node's **`crypto.timingSafeEqual()`**. This is a Node-only API and is **not implemented by `crypto-browserify`** (the standard `crypto` polyfill used by Vite/webpack/most bundlers). As a result, any **browser / Electron-renderer / bundled** consumer throws at runtime when unlocking a keystore:

```
TypeError: crypto.timingSafeEqual is not a function
```

The keystore *format* is unchanged, so this is source-backwards-compatible — but it is **not runtime-compatible for non-Node consumers**.

### Why it slips through CI

Test suites run under Node (Jest/Vitest), where `crypto.timingSafeEqual` exists. The failure only appears in a real browser/renderer bundle, so green CI does not catch it.

### Affected code

`packages/xchain-crypto/src/crypto.ts`, in `decryptFromKeystore`:

```ts
if (computedMac.length !== expectedMac.length || !crypto.timingSafeEqual(computedMac, expectedMac))
  throw new Error('Invalid password')
```

### Who is affected

Any browser-bundled wallet built on xchainjs (desktop Electron renderers, browser extensions, web wallets) using a `crypto` polyfill without `timingSafeEqual` — i.e. essentially all of them, since `crypto-browserify` lacks it.

### Suggested fix

Guard the call and fall back to a manual constant-time comparison when `timingSafeEqual` is unavailable:

```ts
const constantTimeEqual = (a: Buffer, b: Buffer): boolean => {
  if (a.length !== b.length) return false
  if (typeof crypto.timingSafeEqual === 'function') return crypto.timingSafeEqual(a, b)
  // Fallback for browser polyfills (crypto-browserify) that lack timingSafeEqual
  let diff = 0
  for (let i = 0; i < a.length; i++) diff |= a[i] ^ b[i]
  return diff === 0
}
```

This keeps the timing-attack hardening on Node while restoring browser compatibility.

### Workaround (consumer side)

Alias `crypto` to a shim that augments `crypto-browserify` with a `timingSafeEqual` implementation.

### Environment

- `@xchainjs/xchain-crypto` 1.0.7
- Renderer bundler: Vite (electron-vite), `crypto` aliased to `crypto-browserify`

Issue #1721, opened 2026-06-30
xchain-crypto: keystore MAC does not authenticate the IV (CTR-mode malleability)


underthesun49:
### Summary

In `@xchainjs/xchain-crypto`, the BLAKE2b MAC over the keystore is computed from `macKey || ciphertext` but **does not include the IV** (`cipherparams.iv`). Since the cipher is CTR mode, the IV directly determines the keystream offset; an attacker who tampers with the stored IV can flip bits in the decrypted plaintext without invalidating the MAC. Including the IV in the MAC input is standard authenticated-encryption hygiene.

### Where

`packages/xchain-crypto/src/crypto.ts` (observed in published `lib/index.js`, v1.0.7):

```js
// encrypt
const mac_bytes = blake2b(Buffer.concat([derivedKey.slice(16, 32), Buffer.from(cipherText)]), { dkLen: 32 })

// decrypt
const mac_bytes = blake2b(Buffer.concat([derivedKey.slice(16, 32), ciphertext]), { dkLen: 32 })
```

The `iv` (`crypto.randomBytes(16)`, stored as `cipherparams.iv`) is never fed into `blake2b`.

### Suggestion

Include the IV in the MAC input on both encrypt and decrypt, e.g.:

```js
blake2b(Buffer.concat([derivedKey.slice(16, 32), iv, ciphertext]), { dkLen: 32 })
```

### Backwards compatibility

This changes the MAC for all keystores, so it would need to be gated behind a keystore `version` bump (new files include the IV in the MAC; legacy files verified with the current `macKey || ciphertext` construction). Pairs naturally with a versioned-keystore change.

### Impact

Low severity in practice for BIP39 mnemonics — a flipped plaintext almost always fails BIP39 checksum/word validation — but the MAC should authenticate everything an attacker can modify (IV included) as a matter of correct AE construction. Happy to send a PR if a versioned approach is welcome.

PR #1723, opened 2026-07-08
feat(crypto): upgrade keystore encryption to AES-256-CTR (v2)

### LINKED ISSUE #1720: xchain-crypto: keystore uses aes-128-ctr (only 16 of 32 derived bytes); consider aes-256-ctr
### Summary

In `@xchainjs/xchain-crypto`, the keystore encryption derives a 32-byte key via PBKDF2 (`dklen = 32`) but only the first 16 bytes are used for the cipher, and the cipher is hard-coded to `aes-128-ctr`. Bumping to `aes-256-ctr` and feeding the full 32-byte derived key would raise the key-material strength to the modern standard at no extra cost (the bytes are already derived).

### Where

`packages/xchain-crypto/src/crypto.ts` (observed in published `lib/index.js`, v1.0.7):

```js
const cipher = 'aes-128-ctr' // Encryption cipher
const dklen = 32             // Derived key length

// encrypt
const derivedKey = await pbkdf2Async(Buffer.from(password), salt, kdfParams.c, kdfParams.dklen, hashFunction)
const cipherIV = crypto.createCipheriv(cipher, derivedKey.slice(0, 16), iv) // only first 16 bytes used
// derivedKey.slice(16, 32) is reused as the MAC key
```

So of the 32 derived bytes, bytes `0..16` are the AES-128 key and bytes `16..32` are the MAC key.

### Suggestion

Switch the cipher to `aes-256-ctr` and use a full 32-byte AES key (deriving a separate MAC key, e.g. by increasing `dklen` to 64 and splitting 32/32, or deriving the MAC key independently).

### Backwards compatibility

This must be done **without breaking existing keystores**. Decryption already reads the cipher name from `keystore.crypto.cipher`, so old `aes-128-ctr` files can still be decrypted as long as the key-slicing for legacy files is preserved. The cleanest path is to bump the keystore `version` and branch decryption on it: new files use `aes-256-ctr` + full key, old files keep the current 16-byte slice. Happy to open a PR along these lines if the approach sounds good.

### Impact

Low severity — AES-128 has no known practical break — but for software protecting seed phrases, AES-256 is the expected standard and the derived bytes are already available.

betelthyme:
## Summary

Fixes #1720.

Upgrades `@xchainjs/xchain-crypto` keystore encryption from **AES-128-CTR** to **AES-256-CTR**, introducing keystore format **version 2**. The previous scheme derived 32 bytes via PBKDF2 but used only the first 16 as the AES key (`aes-128-ctr`), with bytes 16–32 as the MAC key. We now derive 64 bytes and split them into an **independent** 32-byte AES-256 key and 32-byte MAC key.

## What changed (`src/crypto.ts`)

- Constants: `cipher` `aes-128-ctr` → `aes-256-ctr`; `dklen` `32` → `64`; new `keystoreVersion = 2`.
- New `aesKeyLengthForCipher(cipherName)` helper — maps cipher → AES key length (16 for aes-128, 32 for aes-256), throws on unsupported ciphers.
- `encryptToKeyStore`: splits the derived key into `encryptionKey = slice(0, keyLen)` and `macKey = slice(keyLen, dklen)`, and stamps `version: 2`.
- `decryptFromKeystore`: derives the same split **from the keystore's own `cipher`/`dklen` fields**, so the byte layout is data-driven rather than hard-coded.

## Notable design decisions

- **Independent MAC key (not key reuse).** With a 32-byte AES-256 key, naively reusing `slice(16,32)` as the MAC key would make the MAC key a substring of the encryption key — key reuse across two primitives. Instead `dklen` is raised to 64 so the AES key (`0–32`) and MAC key (`32–64`) never overlap.
- **Source of truth for the key split is `cipher`/`dklen`, not `version`.** Those fields literally determine the byte layout and are already stored in every keystore; `version` is the coarse label. This is what keeps decryption of both formats correct.

## Backward compatibility

- ✅ **Reading old files:** existing v1 (`aes-128-ctr`, `dklen` 32) keystores decrypt unchanged — covered by tests. No migration needed for existing wallets.
- ⚠️ **Writing new files (forward-incompatible):** keystores newly created by this version are v2 and **cannot be opened by older `xchain-crypto` releases** (old code hard-codes a 16-byte key slice and throws on a 32-byte AES-256 key). Existing files are untouched. Integrators sharing keystore files across components on different `xchain-crypto` versions should upgrade them together before creating new keystores. Called out in the changeset.

## Tests (`__tests__/crypto.test.ts`)

- New **v2 known-answer test** — deterministic salt/iv produce a locked-in ciphertext + MAC (locks the v2 format).
- New **v2 decrypt** test.
- Preserved **v1 backward-compat** decrypt fixtures (both `aes-128-ctr` c=600000 and the older c=262144).
- `Export Keystore` test updated to assert `aes-256-ctr` / `dklen 64` / `version 2`.

All 14 tests pass; build, `tsc --noEmit`, and eslint are clean.

## Security / compatibility notes

- AES-128 had no known practical break — this is **modernization to the expected standard** for seed-phrase storage, not a fix for an exploitable weakness (low severity, as noted in #1720).
- `aes-256-ctr` is supported by `crypto-browserify`, so browser/Electron-renderer bundles are unaffected (unlike the recent `timingSafeEqual` regression in #1715).

## Changeset

`minor` bump for `@xchainjs/xchain-crypto`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added support for a newer, stronger keystore format using AES-256 encryption.
  * Newly created keystores now use an updated version format.

* **Bug Fixes**
  * Existing keystores from the previous format remain readable.
  * Import/export behavior now correctly handles both older and newer keystore formats.

### DIFF
--- a/.changeset/crypto-aes256-keystore.md
+++ b/.changeset/crypto-aes256-keystore.md
@@ -1,0 +1,13 @@
+---
+'@xchainjs/xchain-crypto': minor
+---
+
+Upgrade keystore encryption to AES-256-CTR (keystore format v2).
+
+`encryptToKeyStore` now derives a 64-byte key via PBKDF2 (`dklen` 32 → 64) and splits it into an independent 32-byte AES-256 key and 32-byte MAC key, writing `cipher: "aes-256-ctr"` and `version: 2`. Previously it used `aes-128-ctr` with only the first 16 of 32 derived bytes as the AES key (the remaining 16 were the MAC key).
+
+**Backward compatible (reading):** `decryptFromKeystore` derives the AES-key/MAC-key split from the keystore's own `cipher` and `dklen` fields, so existing v1 (`aes-128-ctr`, `dklen` 32) keystores continue to decrypt unchanged. No user action or migration is required to keep opening existing wallets.
+
+**Forward-incompatible (writing) — note for integrators:** keystore files newly created by this version are `aes-256-ctr` / v2 and **cannot be opened by older `@xchainjs/xchain-crypto` releases** (older code hard-codes a 16-byte key slice and will throw on the 32-byte AES-256 key). Existing files are untouched and remain openable everywhere. If your app shares keystore files across components pinned to different `xchain-crypto` versions, upgrade them together before creating new keystores.
+
+Security note: AES-128 had no known practical break, so this is a modernization to the expected standard for seed-phrase storage rather than a fix for an exploitable weakness. The AES-256-CTR path is also supported by `crypto-browserify`, so browser/Electron bundles are unaffected.
--- a/packages/xchain-crypto/__tests__/crypto.test.ts
+++ b/packages/xchain-crypto/__tests__/crypto.test.ts
@@ -18,6 +18,30 @@ describe('Keystore regression test for encrypt/decrypt with internal migration',
   const phrase = 'patient use either flash couple jump castle true broccoli cancel brand mechanic'
   const password = '1234'
 
+  // v2 keystore (aes-256-ctr, dklen 64) — the format written by the current encryptToKeyStore.
+  // Known-answer test: produced from the fixed salt/iv/password/phrase above.
+  const expectedKeystoreV2 = {
+    crypto: {
+      cipher: 'aes-256-ctr',
+      ciphertext:
+        '31607fd3f65a69ded1701f2f45f1bc7eba8d8ac3eb3a9c50e95ac9d5915631e34c59ba05047bfe4f1ae5639081ea705eef5e65fea253b39faa5e2c0e8b606c60606cb1e5c4081518ca4dcb472fe7d9',
+      cipherparams: { iv: 'dffdb8bbe92e9a00e173eaa20f1a3784' },
+      kdf: 'pbkdf2',
+      kdfparams: {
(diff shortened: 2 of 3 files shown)


Thorian1te:
### Security take: not urgent

**v1 is not practically “hackable” relative to v2 today.** AES-128-CTR has no known practical break for this use case; the real offline risk for both formats is a **weak password + stolen keystore file**, not AES width. v2 mainly adds long-term/quantum margin and ~2× PBKDF2 cost per guess (`dklen` 64 vs 32). Shared gaps (e.g. MAC not binding the IV — #1721) affect both.

| Defer / skip | Still worth merging later |
|--------------|---------------------------|
| No live AES-128 break | AES-256 is the expected standard for seed storage |
| Dominant risk is password / device compromise | Modest extra offline-crack cost + future margin |
| Forward-incompatible **writes** (old clients can’t open new v2 keystores) | Work is already done + tested |

**Bottom line:** safe to wait if prioritization is tight — this is modernization/hygiene, not a must-ship patch. If/when it ships, call out the integrator note: upgrade all components that share keystore files before creating new ones.

PR #1740, opened 2026-08-04
fix(cosmos-sdk): preserve real errors in round-robin RPC failover

### LINKED ISSUE #1727: Misleading "No clients available" errors mask real RPC/chain failures (round-robin swallows exceptions)
## Summary

Several round-robin helpers in the Cosmos / THOR / MAYA clients catch **all** errors (including successful connections that return chain rejections or HTTP 5xx) and then throw a generic:

```text
No clients available. Can not …
```

That message is only accurate when **no RPC URL can be used**. In practice it also appears when:

- The RPC is up and returns **HTTP 500** (e.g. Liquify gateway / Tendermint)
- `broadcast_tx` / `tx_search` fail after a partial success
- Chain rejects the tx (**insufficient funds**, bad sequence, etc.)
- `getTx` returns null because the tx is not indexed yet

The real error is often only visible via `console.warn` (deposit path) or discarded entirely (empty `catch`).

## Affected packages / methods

Observed in published builds used by Asgardex (`@xchainjs/xchain-thorchain@3.0.x`, `@xchainjs/xchain-cosmos-sdk@2.0.x`, same pattern in mayachain / cosmos / kujira):

| Location | Generic throw |
|---|---|
| `xchain-thorchain` `roundRobinSignAndBroadcastDeposit` | `No clients available. Can not sign and broadcast deposit transaction` |
| `xchain-thorchain` `roundRobinSign` / `roundRobinSignAndBroadcastTx` | sign / sign-and-broadcast variants |
| `xchain-cosmos-sdk` `roundRobinBroadcast` | `No clients available. Can not broadcast transaction` |
| `xchain-cosmos-sdk` `roundRobinGetTransaction` | `No clients available. Can not retrieve transaction ${txId}` |
| `xchain-cosmos-sdk` `roundRobinGetAccount` / balances / block / chain id | similar |

### Example (deposit)

```js
// xchain-thorchain – deposit broadcast
for (const url of this.clientUrls[this.network]) {
  try {
    // connect + signAndBroadcast
  } catch (e) {
    console.warn(e instanceof Error ? e.message : String(e)) // real error only here
  }
}
throw Error('No clients available. Can not sign and broadcast deposit transaction')
```

Reproduced against a live RPC: CosmJS reported e.g. insufficient funds / fee issues, but callers only received the generic “No clients available” string.

### Example (Ledger / shared broadcast)

```js
// xchain-cosmos-sdk – roundRobinBroadcast
catch (_a) { }  // real error discarded
throw Error(`No clients available. Can not broadcast transaction`)
```

### Example (tx status after send)

```js
// roundRobinGetTransaction
const tx = yield client.getTx(txId)
if (!tx) throw Error(`Can not find transaction ${txId}`)
// …
throw Error(`No clients available. Can not retrieve transaction ${txId}`)
```

HTTP **500** on `tx_search` / `getTx` is also collapsed into this message when the empty catch is used.

## Why this matters for consumers

Wallets (e.g. Asgardex Desktop) surface `error.message` in the UI. Users see **“No clients available”** and assume misconfiguration, while the actual problem is:

- Provider rate limit / gateway **500**
- Insufficient RUNE for native fee
- Tx not yet indexed (status poll)

Debugging is harder because renderer Network may not show main-process Ledger traffic, and the only real signal is a swallowed exception.

## Suggested improvement

1. **Do not treat all failures as “try next client”.**
   - Continue round-robin only on **connection / transport** failures (DNS, timeout, network down, maybe 429/503 with backoff).
   - On **application / chain** failures (DeliverTx code ≠ 0, insufficient funds, invalid tx) or definitive HTTP errors after a connected client, **rethrow the original error** (or stop with that error).

2. **If all endpoints fail, aggregate context:**
   ```text
   No clients available. Can not broadcast transaction.
   Last error: Bad status on response: 500
   ```
   or `cause: lastError` on a custom error class.

3. **For `getTx` / not-found:** use a distinct error (e.g. `TxNotFoundError`) so status pollers can retry without treating it as “no RPC clients”.

4. Apply consistently across thorchain, mayachain, cosmos-sdk, cosmos, kujira round-robin helpers.

## Environment / notes

- CosmJS `@cosmjs/stargate` / `@cosmjs/tendermint-rpc` already surface useful messages (e.g. `Bad status on response: 500`, codespace logs).
- Keystore paths that don’t wrap in try/catch already show the real CosmJS error; deposit / `broadcastTx` / `getTransaction` paths that catch-all do not — inconsistent UX for the same RPC issue.
- Authenticated provider URLs (e.g. Liquify portal `https://gateway.liquify.com/api=<KEY>`) work fine as plain `clientUrls` strings; no change needed for path-style API keys. This issue is only about **error fidelity** in round-robin.

## Acceptance criteria

- [ ] Real CosmJS / chain errors are preserved (or attached as `cause` / “last error”) when round-robin fails.
- [ ] “No clients available” is reserved for cases where no endpoint could be used (empty `clientUrls` or all connection failures).
- [ ] Tx-not-found is distinguishable from client/RPC unavailability for status polling.
- [ ] Same behavior on THOR deposit, MsgSend broadcast, and `getTransactionData`.

Happy to help test against Asgardex Desktop + public/authenticated Liquify RPC if useful.

Thorian1te:
## Summary

Fixes #1727 — Cosmos/THOR/MAYA round-robin helpers no longer collapse every failure into a generic **“No clients available”** message.

- Add shared `roundRobinTry`, `isRetryableRpcError`, `TxNotFoundError`, and exhausted-error helpers in `@xchainjs/xchain-cosmos-sdk`
- **Non-retryable** chain errors (e.g. insufficient funds, sequence mismatch) rethrow immediately
- **Retryable** transport/gateway failures still try the next URL
- When all endpoints fail, the final error includes **Last error: …** and `error.cause`
- `getTransaction` throws **`TxNotFoundError`** when reachable nodes report the tx as missing (distinct from RPC unavailability)

Applied consistently in cosmos-sdk, thorchain, mayachain, cosmos (keystore + ledger), and kujira.

## Test plan

- [x] Unit tests: `packages/xchain-cosmos-sdk/__tests__/roundRobin.test.ts` (9 tests)
- [x] Build affected packages
- [ ] Asgardex / suite: insufficient-funds deposit shows real CosmJS message
- [ ] All RPCs returning 500 surfaces last error in the exhausted message
- [ ] Tx status poll: missing tx is `TxNotFoundError`, not “no clients”

## Summary by CodeRabbit

- **New Features**
  - Added consistent automatic failover across available RPC endpoints for Cosmos transactions and queries.
  - Improved handling of retryable and non-retryable RPC errors.
  - Clearer errors are now reported when all endpoints fail, including the underlying cause.
  - Missing transactions now return a specific transaction-not-found error.
  - Exposed reusable round-robin retry and error-handling utilities.

- **Bug Fixes**
  - Prevented chain and application errors from being incorrectly retried or silently discarded.
  - Improved retry behavior for signing, broadcasting, and deposit operations across Cosmos-based chains.

### DIFF
--- a/.changeset/cosmos-round-robin-error-fidelity.md
+++ b/.changeset/cosmos-round-robin-error-fidelity.md
@@ -1,0 +1,9 @@
+---
+'@xchainjs/xchain-cosmos-sdk': patch
+'@xchainjs/xchain-thorchain': patch
+'@xchainjs/xchain-mayachain': patch
+'@xchainjs/xchain-cosmos': patch
+'@xchainjs/xchain-kujira': patch
+---
+
+Preserve real RPC/chain errors in Cosmos round-robin helpers instead of always throwing a generic "No clients available" message. Non-retryable failures (e.g. insufficient funds) rethrow immediately; exhausted failover includes the last error as message + cause. `getTransaction` uses `TxNotFoundError` when the tx is missing on reachable nodes.
--- a/packages/xchain-cosmos-sdk/__tests__/roundRobin.test.ts
+++ b/packages/xchain-cosmos-sdk/__tests__/roundRobin.test.ts
@@ -1,0 +1,80 @@
+import {
+  TxNotFoundError,
+  createRoundRobinExhaustedError,
+  isRetryableRpcError,
+  roundRobinTry,
+} from '../src/roundRobin'
+
+describe('isRetryableRpcError', () => {
+  it('treats connection / gateway failures as retryable', () => {
+    expect(isRetryableRpcError(new Error('Failed to fetch'))).toBe(true)
+    expect(isRetryableRpcError(new Error('connect ECONNREFUSED 127.0.0.1:26657'))).toBe(true)
+    expect(isRetryableRpcError(new Error('Bad status on response: 500'))).toBe(true)
+    expect(isRetryableRpcError(new Error('Bad status on response: 503'))).toBe(true)
+    expect(isRetryableRpcError(new Error('request timed out'))).toBe(true)
+  })
+
+  it('treats chain rejection messages as non-retryable', () => {
+    expect(isRetryableRpcError(new Error('insufficient funds: insufficient account funds'))).toBe(false)
+    expect(isRetryableRpcError(new Error('account sequence mismatch, expected 5, got 4'))).toBe(false)
+    expect(isRetryableRpcError(new Error('out of gas in location: WritePerByte'))).toBe(false)
+    expect(isRetryableRpcError(new Error('failed to execute message; message index: 0'))).toBe(false)
+  })
+})
+
+describe('createRoundRobinExhaustedError', () => {
+  it('includes last error message and cause', () => {
+    const cause = new Error('Bad status on response: 500')
+    const err = createRoundRobinExhaustedError('broadcast transaction', cause)
+    expect(err.message).toContain('No clients available. Can not broadcast transaction')
+    expect(err.message).toContain('Last error: Bad status on response: 500')
+    expect((err as Error & { cause?: unknown }).cause).toBe(cause)
+  })
+
+  it('works without a last error', () => {
+    const err = createRoundRobinExhaustedError('get chain id')
+    expect(err.message).toBe('No clients available. Can not get chain id')
+  })
+})
+
+describe('roundRobinTry', () => {
+  it('returns the first successful result', async () => {
+    const result = await roundRobinTry([1, 2, 3], 'test', async (n) => {
+      if (n < 3) throw new Error('Failed to fetch')
+      return n * 10
+    })
+    expect(result).toBe(30)
+  })
+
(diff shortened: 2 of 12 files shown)


Thorian1te:
Addressed CodeRabbit feedback in latest commit:

1. **TxNotFound** — only when *every* client returns `null` (`roundRobinGetTx`). Mixed null + provider failure → exhausted error with last cause. Regression tests for both orderings.
2. **Retry classifier** — positive allowlist for connection/DNS/socket/timeout and HTTP 429/5xx only. 4xx and unknown messages are non-retryable.
3. **sign-once / broadcast-only retry** — acknowledged as a real double-submit risk, but it predates this PR (old empty-catch loops already re-ran `signAndBroadcast`). Tracking as a follow-up rather than expanding this error-fidelity change.

underthesun49 (approved):

PR #1749, opened 2026-08-12
fix(crypto): include CTR IV in keystore MAC (v2)

### LINKED ISSUE #1721: xchain-crypto: keystore MAC does not authenticate the IV (CTR-mode malleability)

Thorian1te:
## Summary

Fixes #1721.

Keystore MAC previously covered only `macKey || ciphertext`, not the CTR **IV**. An attacker who could modify `cipherparams.iv` could change the decrypted plaintext without invalidating the MAC (CTR malleability). Low practical impact for BIP39 (checksum usually fails), but the MAC should authenticate everything an attacker can edit.

### Change

| | v1 (legacy, still readable) | v2 (new writes) |
|--|----------------------------|-----------------|
| MAC input | `macKey \|\| ciphertext` | `macKey \|\| iv \|\| ciphertext` |
| `version` | `1` | `2` |
| Cipher | unchanged (`aes-128-ctr`) | unchanged |

- **Encrypt:** always writes `version: 2` with IV bound into the MAC.
- **Decrypt:** `version >= 2` → verify with IV; `version < 2` → legacy MAC (existing wallets keep working).
- **Tests:** known-answer v2 vector; v1 + old PBKDF2 iteration fixtures still decrypt; IV tamper fails on v2; documents that v1 still accepts IV tamper (legacy gap).

### Not in this PR

- AES-256 / independent 32-byte keys — tracked by #1720 / open #1723. When that lands, please **keep IV-in-MAC** for new versions (rebase onto this or fold the same MAC rule).

### Compatibility

- **Reading old keystores:** yes (v1 MAC path retained).
- **Older library reading new keystores:** no — v2 MAC will not match old verifiers. Integrators that share keystore files across mixed `xchain-crypto` versions should upgrade together before creating new keystores.

## Test plan

- [x] `yarn workspace @xchainjs/xchain-crypto test`
- [ ] Reviewer: confirm version gate (`>= 2`) is the right switch vs cipher-based
- [ ] Note interaction with #1723 (AES-256 v2) for sequencing

## Summary by CodeRabbit

* **Security Enhancements**
  * New keystores use an updated format that authenticates the initialization vector, helping detect IV tampering.
  * Existing version 1 keystores remain readable.
  * Incorrect passwords and modified keystore data continue to be rejected safely.
* **Compatibility**
  * Version 2 keystores require a release that supports the updated format.

### DIFF
--- a/.changeset/crypto-keystore-mac-iv.md
+++ b/.changeset/crypto-keystore-mac-iv.md
@@ -1,0 +1,17 @@
+---
+'@xchainjs/xchain-crypto': minor
+---
+
+Include the CTR IV in the keystore MAC (format version 2)
+
+New keystores written by `encryptToKeyStore` use `version: 2` and compute
+`blake2b(macKey || iv || ciphertext)` so an attacker cannot flip bits in the
+decrypted plaintext by tampering with `cipherparams.iv` without invalidating
+the MAC (#1721).
+
+**Reading:** v1 keystores (`version: 1`, MAC over `macKey || ciphertext` only)
+still decrypt unchanged.
+
+**Writing:** newly created files are v2 and cannot be verified by older
+`@xchainjs/xchain-crypto` releases that only check the v1 MAC layout. Existing
+files are unaffected.
--- a/packages/xchain-crypto/__tests__/crypto.test.ts
+++ b/packages/xchain-crypto/__tests__/crypto.test.ts
@@ -18,7 +18,8 @@ describe('Keystore regression test for encrypt/decrypt with internal migration',
   const phrase = 'patient use either flash couple jump castle true broccoli cancel brand mechanic'
   const password = '1234'
 
-  const expectedKeystore = {
+  // v1 fixture: MAC = blake2b(macKey || ciphertext) — IV not covered (#1721 legacy)
+  const expectedKeystoreV1 = {
     crypto: {
       cipher: 'aes-128-ctr',
       ciphertext:
@@ -35,6 +36,18 @@ describe('Keystore regression test for encrypt/decrypt with internal migration',
     },
     id: '9ad9ea91-22ad-46a7-9613-4f9d190e32ab',
     version: 1,
+    meta: 'xchain-keystore',
+  }
+
+  // Same salt/iv/ciphertext as v1, but MAC includes IV (v2). Ciphertext is unchanged because
+  // AES-CTR still uses the same key material layout; only the MAC binding changes.
+  const expectedKeystoreV2 = {
+    crypto: {
+      ...expectedKeystoreV1.crypto,
+      mac: '801caafeca2863d0486c00464fb954ff512e8146e331710173f2c2994068f89f',
+    },
+    id: '9ad9ea91-22ad-46a7-9613-4f9d190e32ab',
+    version: 2,
     meta: 'xchain-keystore',
   }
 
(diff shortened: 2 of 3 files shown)


Thorian1te:
## Practical risk scale for #1721 (for reviewers)

Assuming the question is “how likely is an attacker to actually use CTR IV not being in the MAC?” — put it on a scale:

### Overall practical risk: **~1–2 / 10** (low)

| Dimension | Score | Why |
|-----------|--------|-----|
| **Crypto correctness** | **7/10** “should fix” | Real AE gap: MAC doesn’t bind IV; textbook CTR malleability |
| **Exploitability in the wild** | **1–2/10** | Needs write access to the keystore file **and** a useful plaintext goal |
| **Impact if fully exploited** | **3–4/10** for BIP39 wallets | Garbled mnemonic almost always fails BIP39 wordlist/checksum → decrypt “succeeds” as garbage or fails later |
| **Impact if non-BIP39 / raw secrets** | **6–7/10** | If you ever stored arbitrary secrets this way, bit flips matter more |
| **Priority vs other crypto work** | Below wrong password handling, seed theft, XSS, malicious extension | Those actually steal funds |

The issue text already says this: **low severity in practice for BIP39**.

### What an attacker would need

1. **Modify** the stored keystore (disk, backup, synced config, compromised device, malicious backup provider).
2. Change **`cipherparams.iv`** (and leave ciphertext + MAC as-is).
3. Victim opens wallet with the **correct password**.
4. MAC still passes (v1), CTR decrypts with wrong keystream → **wrong bytes**, not the real seed “extracted cleanly.”

They do **not** get the password or a clean copy of the mnemonic from this alone. They get **integrity failure that looks like “valid password, bad data”** (or a non-mnemonic string).

### Likelihood of real-world abuse

| Scenario | Chance someone “uses” this | Notes |
|----------|----------------------------|--------|
| Random internet attacker, remote | **~0** | No keystore write path |
| Malware already on the machine | **Low value** | If they can edit keystore, they can often steal it + wait for password, or replace whole file, or keylog — better attacks |
| Evil backup / cloud sync | **Low** | Same: full file replace or ransomware is simpler than surgical IV flips |
| Targeted crypto nerd demo | **Yes, in a lab** | Easy to show MAC passes + plaintext changes |
| Turning flipped bits into a **chosen valid BIP39** | **Very hard** | BIP39 checksum + wordlist; not free bit-control to a funded wallet |

So: **great for a security write-up / “fix the construction”; poor as a primary theft primitive.**

### Scale analogy

```
0 ──────── 1–2 ──── 3–4 ──────── 6–7 ──────── 9–10
|           |        |            |             |
|        #1721    serious      seed in       active
|        practical  bug if      plaintext    RCE +
|        BIP39      non-BIP39   / bad KDF     wallet drain
|        (here)
```

- **Chance an attacker uses this for real theft:** about **1–2/10**.
- **Should we still fix it?** **Yes (~8/10 for hygiene)** — cheap, versioned, correct crypto; that’s why this PR exists.

### Compared to things that *are* higher risk

| Issue class | Relative risk |
|-------------|----------------|
| Malicious extension / clipboard / phishing | Much higher |
| Keystore file + password phishing | Higher |
| Weak password + stolen keystore | Higher |
| #1721 IV not in MAC | **Low** practical |
| AES-128 vs 256 (#1720) | Also mostly hygiene at this threat model |

**Bottom line:** Treat #1721 as **real but niche** — fix for correctness (this PR), don’t rank it as “attackers are probably doing this.” On a 0–10 “will this get used against users” scale: **about 1–2**.

